PR #3, opened 2017-04-30
added Python 3 compatibility for the `reduce` function

roxyboy:
The `reduce` function seems to have moved under the `functools` module for Python 3.

### DIFF
--- a/xrft/xrft.py
+++ b/xrft/xrft.py
@@ -1,6 +1,7 @@
 import numpy as np
 import xarray as xr
 import pandas as pd
+import functools as ft
 import dask.array as dsar
 
 __all__ = ["dft"]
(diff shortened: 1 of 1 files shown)


xrft/xrft.py
```diff
@@ -77,7 +78,10 @@ def dft(da, dim=None, shift=True, remove_mean=True, density=False):
     dk = [l[1] - l[0] for l in k]
 
     if density:
-        spectral_volume = reduce(lambda x, y: x*y, dk)
+        try:
+            spectral_volume = reduce(lambda x, y: x*y, dk)
+        except:
+            spectral_volume = ft.reduce(lambda x, y: x*y, dk)
```
rabernat: let's just use functools always! simpler

rabernat: Can you make this little change this so I can merge it?

PR #4, opened 2017-05-03
Isotropic

roxyboy:
added isotropic PSD calculation in `xrft.py` and example notebook under scripts/


xrft/xrft.py
```diff
@@ -77,13 +83,36 @@ def dft(da, dim=None, shift=True, remove_mean=True, density=False):
     dk = [l[1] - l[0] for l in k]
 
     if density:
-        spectral_volume = reduce(lambda x, y: x*y, dk)
+        try:
+            spectral_volume = reduce(lambda x, y: x*y, dk)
+        except:
+            spectral_volume = ft.reduce(lambda x, y: x*y, dk)
```
rabernat: let's just use functools for the future

xrft/xrft.py
```diff
@@ -77,13 +83,36 @@ def dft(da, dim=None, shift=True, remove_mean=True, density=False):
...
+            invalid = kidx[-1]
+            area = np.bincount(kidx)
+
+            kr = np.ma.masked_invalid(np.bincount(kidx,
+                                                weights=K.ravel()) / area)
+            iso_f = np.ma.masked_invalid(np.bincount(kidx,
+                                                    weights=np.real(f
+                                                    *np.conj(f)).ravel())
+                                                    / area) * kr
+
```
rabernat: Just one though about the structure of this... Might it not be better to have a separate function called `power_spectrum` that does higher level stuff?

rabernat:
just to clarify, I think we should have a few new functions in xrft.py: 

  - `power_spectrum(da, density=True/False)` which would call `dft` internally
  - `cross_spectrum(da1, da2)` 
  - `isotropic_power_spectrum(da)`, which would call `power_spectrum` internally
  - `isotropic_cross_spectrum(da1, da2)

The later two functions would call an internal (non-public) function called `_azimuthal_average`, which would include the code in your current PR.

xrft/xrft.py
```diff
@@ -96,7 +91,288 @@ def dft(da, dim=None, shift=True, remove_mean=True, density=False):
...
+            if pd.core.common.is_timedelta64_dtype(diff):
+                # convert to seconds so we get hertz
+                diff = diff.astype('timedelta64[s]').astype('f8')
+            delta = diff[0]
+            delta_x.append(delta)
+        np.testing.assert_almost_equal(ps.sum()
+                                       / (np.asarray(delta_x).prod()
+                                          * (da.values**2).sum()
+                                         ), 1., decimal=5
+                                      )
```
rabernat: It seems like this test should be in the tests, not the function itself, no?

xrft/xrft.py
```diff
@@ -96,7 +91,288 @@ def dft(da, dim=None, shift=True, remove_mean=True, density=False):
...
+    Parameters
+    ----------
+    da1 : `xarray.DataArray`
+        The data to be transformed
+    da2 : `xarray.DataArray`
+        The data to be transformed
+    a1 : float64
+        Coefficient of da1
+    a2 : float64
+        Coefficient of da2
```
rabernat: what is the meaning of these coefficients?

xrft/xrft.py
```diff
@@ -96,7 +91,288 @@ def dft(da, dim=None, shift=True, remove_mean=True, density=False):
...
+    Returns
+    -------
+    cs : `xarray.DataArray`
+        Two-dimensional cross spectrum
+    """
+    if dim is None:
+        dim = da1.dims
+        dim2 = da2.dims
+        if len(dim) != len(dim2):
+            raise ValueError('The two datasets have different dimensions')
```
rabernat: What happens if there is the same _number_ of dimensions, but they are not the same dimensions (e.g. ['x', 'y'] vs ['lon', 'lat'] would pass the check).

You may want to consider calling [xarray.align](http://xarray.pydata.org/en/stable/generated/xarray.align.html) on the data.

xrft/xrft.py
```diff
@@ -96,7 +91,288 @@ def dft(da, dim=None, shift=True, remove_mean=True, density=False):
...
+            if pd.core.common.is_timedelta64_dtype(diff):
+                # convert to seconds so we get hertz
+                diff = diff.astype('timedelta64[s]').astype('f8')
+            delta = diff[0]
+            delta_x.append(delta)
+        np.testing.assert_almost_equal(cs.sum()
+                                       / (np.asarray(delta_x).prod()
+                                          * (a1*a2*da1.values*da2.values).sum()
+                                         ), 1., decimal=5
+                                      )
```
rabernat: same comment as above...tests should go in the test module, not the function itself.

xrft/xrft.py
```diff
@@ -96,7 +91,288 @@ def dft(da, dim=None, shift=True, remove_mean=True, density=False):
...
+
+    N = [da1.shape[n] for n in axis_num]
+
+    daft1 = dft(da1, dim=dim,
+                shift=shift, remove_mean=remove_mean)
+    daft2 = dft(da2, dim=dim,
+                shift=shift, remove_mean=remove_mean)
+    coord = list(daft1.coords)
+
+    cs = np.real(a1*daft1 * a2*np.conj(daft2))
```
rabernat: why are a1 and a2 necessary? couldn't the user just pass `a1*da1` and `a2*da2` if he wants to rescale the inputs?

xrft/xrft.py
```diff
@@ -1,12 +1,15 @@
...
+__all__ = ["dft","power_spectrum","cross_spectrum",
+            "isotropic_powerspectrum","isotropic_crossspectrum",
+            "fit_loglog"]
 
 
-def dft(da, dim=None, shift=True, remove_mean=True, density=False):
+def dft(da, dim=None, shift=True, remove_mean=True):
     """
     Perform discrete Fourier transform of xarray data-array `da` along the
     specified dimensions.
```
rabernat: I recommend including the mathematical expression for what is calculated directly in the docstring. See [numpydoc documenation](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) for info on how to do this.

xrft/xrft.py
```diff
@@ -96,7 +91,288 @@ def dft(da, dim=None, shift=True, remove_mean=True, density=False):
...
 
+    dk = [l[1] - l[0] for l in k]
     for this_dk, d in zip(dk, dim):
         newcoords[prefix + d + '_spacing'] = this_dk
 
     return xr.DataArray(f, dims=newdims, coords=newcoords)
+
+def power_spectrum(da, dim=None, shift=True, remove_mean=True, density=True):
+    """
+    Calculates the power spectrum of da.
```
rabernat: I recommend including the mathematical expression for what is calculated directly in the docstring. See [numpydoc documenation](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) for info on how to do this.

xrft/xrft.py
```diff
@@ -96,7 +91,288 @@ def dft(da, dim=None, shift=True, remove_mean=True, density=False):
...
+                                          * (da.values**2).sum()
+                                         ), 1., decimal=5
+                                      )
+
+    return xr.DataArray(ps, coords=daft.coords, dims=daft.dims)
+
+def cross_spectrum(da1, da2, a1=1., a2=1., dim=None,
+                   shift=True, remove_mean=True, density=True):
+    """
+    Calculates the cross spectra of da1 and da2.
```
rabernat: I recommend including the mathematical expression for what is calculated directly in the docstring. See [numpydoc documenation](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) for info on how to do this.

xrft/xrft.py
```diff
@@ -96,7 +91,288 @@ def dft(da, dim=None, shift=True, remove_mean=True, density=False):
...
+                                            weights=f.values.ravel())
+                                            / area
+                                            ) * kr
+    return kr, iso_f
+
+def isotropic_powerspectrum(da, dim=None, shift=True, remove_mean=True,
+                       density=True, nbins=64):
+    """
+    Calculates the isotropic spectrum from the
+    two-dimensional power spectrum.
```
rabernat: I recommend including the mathematical expression for what is calculated directly in the docstring. See [numpydoc documenation](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) for info on how to do this.

xrft/tests/test_xrft.py
```diff
@@ -61,3 +61,106 @@ def test_dft_1d_time():
...
+    """Test the spectral slope of synthetic data."""
+    theta = synthetic_field(N, dL, amp, s)
+    theta = xr.DataArray(theta, dims=['k', 'l'],
+                        coords={'k':range(N), 'l':range(N)})
+    iso_f = xrft.isotropic_powerspectrum(theta, remove_mean=True,
+                  density=True)
+    y_fit, a, b = xrft.fit_loglog(iso_f.freq_r.values[4:],
+                            iso_f.values[4:])
+
+    npt.assert_allclose(a, s, atol=.1)
```
rabernat: This is a good start for the tests. But I think you should have a dedicated test for _each function_ in the xrft module. You should make sure all the optional arguments work and do what they say they do. These are called [unit testing](https://en.wikipedia.org/wiki/Unit_testing)

I know this seems like a pain, but it is really worth it in the long run.

rabernat:
Maybe @dhruvbalwada wants to take a look at this? It will be very useful for his upcoming spectral analysis work.

rabernat:
Also, it looks like the travis build is failing. You will need to figure that out before we can merge.

roxyboy:
Each environment on Travis fail on different tests (Python 2.7 and 3.4 fail on the test_dft_2d_slope, and 3.5 fails on test_parseval), which I don't know why... All the tests pass on my local Python 3 environment which the packages I use are all up to date.

rabernat:
Did you have a look at the log from the Travis CI build? https://travis-https://travis-ci.org/rabernat/xrft/jobs/234148267#L493

Here it is failing when you call
```python
    p = np.polyfit(np.log2(x), np.log2(y), 1)
```

because the second argument has NaNs in it. This argument (`np.log2(y)`) gets set when you call
```python
        y_fit, a, b = xrft.fit_loglog(iso_f.freq_r.values[4:],
                               iso_f.values[4:])
```

so the problem is that `iso_f.values[4:]` has NaNs in int. Going farther upstream to see how `iso_f` get set, you have
```python
         iso_f = xrft.isotropic_powerspectrum(theta, remove_mean=True,
                      density=True)
```
...so that means there is a problem with `isotropic_powerspectrum`.

By adding unit tests for all of the functions (including `isotropic_powerspectrum`), you will help narrow down the source of the problem until we can pinpoint it exactly. So keep updating this pull requests with more tests.

xrft/xrft.py
```diff
@@ -9,11 +9,43 @@
             "fit_loglog"]
 
 
-def dft(da, dim=None, shift=True, remove_mean=True):
+def _hanning(da, N):
+    """Apply Hanning window"""
```
rabernat: Adding windowing is a good idea. But why don't you save this for a separate pull request? This current pull request is already failing the tests. We don't need to complicate it further by adding more (untested) functions. In general, pull requests should contain a single new feature or bug fix and be as [small](http://robertheaton.com/2015/10/26/why-and-how-to-make-smaller-pull-requests/) as possible.

Instead, stash or copy the `hanning` function and put it in a new, separate feature branch and submit as a new PR.

rabernat:
So it looks like 
```python
 ps = xrft.power_spectrum(da, window=True)
```
is giving you something with `inf` in it. But your power spectrum test is working.

Try to add more examples to `test_power_spectrum`, starting with the one from  `test_parseval`. Try to figure out why you are getting `inf`s.

rabernat:
Another way to accelerate finding the issue would be to create a local python 3.5 environment to test in.

roxyboy:
I think there's something unstable about the Travis environment for Python 3. I didn't change the xrft.py code but now it's passing in Python 3.4 environment (for some unknown reason).

rabernat:
It's more likely that the numpy / scipy versions were generating the problem, rather than python 3 itself.

rabernat:
@roxyboy this looks almost ready! I'm glad you sorted out the failing tests!

There are just a few more issues to clean up.

- [ ] Remove the ipython notebooks from your PR...those don't belong in the repository
- [ ] Add some tests for the various errors that you raise. As you can see from the [coverage report](https://codecov.io/gh/rabernat/xrft/compare/e05a90da5410e50ed6defe180ff7cd335c29167a...3987bae4cbc6f881911bfabe8526ae38fd797cde/src/xrft/xrft.py), the big parts of `xrft` which are not covered are related to the raising of exceptions. (You can use the [pytest.raises](https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions) context manager to test for exceptions being raised.)
- [ ] Add a test for `isotropic_crossspectrum`

Once you have these checked off, we can merge.

PR #11, opened 2017-05-26
Tests using dask array work now locally

roxyboy:


rabernat:
This looks great! And coverage is improved.

PR #13, opened 2017-05-26
Tests using dask array work now

roxyboy:


xrft/tests/test_xrft.py
```diff
@@ -82,6 +82,16 @@ def test_dft_2d():
...
+def test_dft_3d():
+    """Test the discrete Fourier transform on 3D dask array data"""
+    N=16
+    da = xr.DataArray(np.random.rand(2,N,N), dims=['time','x','y'],
+                      coords={'time':range(2),'x':range(N),
+                              'y':range(N)}).chunk({'time': 1}
+                     )
+    daft = xrft.dft(da, dim=['x','y'], shift=False, remove_mean=False)
+    npt.assert_almost_equal(daft.values, np.fft.fftn(da.values, axes=[1,2]))
+
```
rabernat: A few suggestions about this test.

- Let's explicitly call it a dask test, i.e. `test_dft_3d_dask`. 
- Before calling `values`, you want to check that `daft.data` is also a dask array. The checks that the calculation is being done "lazily", i.e. not evaluating until you access the data.
- You also want to check for errors raised if the chunking is wrong. So you could cal `da.chunk{'time': 1, 'x': 1}`, and then recall `xrft.dft` within a `pytest.raises` context.

rabernat:
Fantastic!

In a future refactoring, we can create [fixtures](https://docs.pytest.org/en/latest/fixture.html) to eliminate the repetitive code in `test_xrft.py`.

PR #14, opened 2017-05-26
Added tests for dask array applying the Hann window and changed the f…

roxyboy:
I think I'm done with editing this for a while.

### DIFF
--- a/xrft/xrft.py
+++ b/xrft/xrft.py
@@ -290,19 +290,19 @@ def _azimuthal_avg(k, l, f, fftdim, N, nfactor):
                                     weights=f.data.ravel())
                                     / area) * kr
     elif f.ndim == 3:
-        iso_f = np.zeros((M,nbins+1))
+        iso_f = np.zeros((M[0],nbins+1))
         if n_dim == 0:
-            for i in range(M):
+            for i in range(M[0]):
                 iso_f[i] = np.ma.masked_invalid(np.bincount(kidx,
                                                 weights=f.data[i].ravel())
                                                 / area) * kr
         elif n_dim == 1:
-            for i in range(M):
+            for i in range(M[0]):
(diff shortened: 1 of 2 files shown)


rabernat:
It looks like there is a lot of stuff in this PR. Can you summarize what issues it addresses?

roxyboy:
The main change is that I edited the `_azimuthal_avg` function so that it can take 2D, 3D and 4D data (the averaging is still taken upon two dimensions but the user can now choose the two dimensions, e.g. if the data has x,y,z,t coordinates, it can calculate the isotropic spectrum on any of the two dimensions out of the four). 
However, I couldn't make it use the metadata provided by `xarray` so I pushed the arguments based on `numpy`

xrft/xrft.py
```diff
@@ -243,37 +252,100 @@ def cross_spectrum(da1, da2, dim=None,
...
+                    for i in range(M[1]):
+                        iso_f[j,i] = np.ma.masked_invalid(np.bincount(kidx,
+                                                weights=f.data[j,:,i].ravel())
+                                                        / area) * kr
+            else:
+                for j in range(M[0]):
+                    for i in range(M[1]):
+                        iso_f[j,i] = np.ma.masked_invalid(np.bincount(kidx,
+                                                weights=f.data[j,:,:,i].ravel())
+                                                        / area) * kr
```
rabernat: I can't merge this PR as long as it contains this long `if / else`  section.

There must be a better way to generalize the logic of azimuthal averaging to multiple dimensions. You should only have one call to `np.bincount`. The code above is very difficult to understand and would be hard to maintain.

If you would like to move on with the PR without getting bogged down in this issue, just change `_azimuthal_avg` to raise an error if you attempt to call it on a field with more than two dimensions.

PR #15, opened 2017-06-20
Added Linear detrend function for data that has more than one dimension

roxyboy:


xrft/xrft.py
```diff
@@ -38,7 +41,39 @@ def _hanning(da, N):
...
+    for i in range(ny):
+        G[nx*i:nx*i+nx, 0] = i+1
+        G[nx*i:nx*i+nx, 1] = np.arange(1, nx+1)
+    m_est = np.dot(np.dot(lin.inv(np.dot(G.T, G)), G.T), d_obs)
+    d_est = np.dot(G, m_est)
+    lin_trend = np.reshape(d_est, (ny,nx))
+
+    da -= lin_trend
+
+    return da
```
rabernat: Does this function work with 1D data?

xrft/xrft.py
```diff
@@ -95,8 +132,19 @@ def dft(da, dim=None, shift=True, remove_mean=True, window=False):
...
     k = [ np.fft.fftfreq(Nx, dx) for (Nx, dx) in zip(N, delta_x) ]
 
-    if remove_mean:
+    if detrend == 'constant':
         da = da - da.mean(dim=dim)
+    elif detrend == 'linear':
+        if len(axis_num) == 1:
+            da = xr.DataArray(sig.detrend(da.values, axis=axis_num[0],
+                                        type=detrend),
+                            dims=da.dims, coords=da.coords)
```
rabernat: Here you use the scipy function instead of your own function. Why?

How does scipy.signal.detrend handle dask arrays?

rabernat:
This looks pretty good.

I'm just concerned that it might not play well with dask arrays.

xrft/xrft.py
roxyboy: No, so I have a `if` syntax in the `dft` function to differentiate 1D and 2D.

xrft/xrft.py
roxyboy: I played around with `scipy.signal.detrend` a bit but it didn't work when I tried to input more than one axis to detrend over...

roxyboy:
Yea, you were write, `scipy.signal.detrend` doesn't keep the dask attribute

xrft/xrft.py
rabernat: Try wrapping this with map_blocks

roxyboy:
I think the `detrend` function works now for dask arrays but the output of `dask.array.map_blocks` returns a numpy array so if the data is detrended, the `fft` will be taken over an `xarray.DataArray` (for now).

rabernat:
Ok this is fine for now.

We will have to clean up and refactor this a bit at some point over the summer.

Issue #16, opened 2017-07-16
Problem with windowing


jbusecke:
Hey everybody, 

just trying to get started with xrft. I noticed an error when I try to apply a window to Aviso ssh fields.

I was able to reproduce the error with a dummy dataset (see below). But in the current state I was not able to get the windowing running.

```
import xrft
import xarray as xr
import numpy as np

data = xr.DataArray(np.random.random([20,30,100]),dims=['time','lat','lon'],coords={'time':range(20),'lat':range(30),'lon':range(100)})
ps = xrft.power_spectrum(data,dim=['time'],window=True)
```

This produces the following error for me:
```
IndexErrorTraceback (most recent call last)
<ipython-input-12-906860386673> in <module>()
     10 
     11 data = xr.DataArray(np.random.random([20,30,100]),dims=['time','lat','lon'],coords={'time':range(20),'lat':range(30),'lon':range(100)})
---> 12 ps = xrft.power_spectrum(data,dim=['time'],window=True)
     13 ps

/home/julius/code/conda_envs/src/xrft/xrft/xrft.pyc in power_spectrum(da, dim, shift, detrend, density, window)
    273     daft = dft(da,
    274             dim=dim, shift=shift, detrend=detrend,
--> 275             window=window)
    276 
    277     coord = list(daft.coords)

/home/julius/code/conda_envs/src/xrft/xrft/xrft.pyc in dft(da, dim, shift, detrend, window)
    189 
    190     if window:
--> 191         da = _hanning(da, N)
    192 
    193     # the hard work

/home/julius/code/conda_envs/src/xrft/xrft/xrft.pyc in _hanning(da, N)
     17     """Apply Hanning window"""
     18 
---> 19     window = np.hanning(N[-1]) * np.hanning(N[-2])[:, np.newaxis]
     20 
     21     # dim = da.dims

IndexError: list index out of range
```

roxyboy:
The windowing at the moment is only expected to take the two horizontal spatial dimensions, but I can see how it would be nice to be able to apply it to one-dimensional data.

PR #18, opened 2017-07-18
Added options for windowing

roxyboy:
It takes into account the issue raised in #16.

### DIFF
--- a/xrft/xrft.py
+++ b/xrft/xrft.py
@@ -13,10 +13,32 @@ __all__ = ["detrend2","_detrend_wrap",
             "fit_loglog"]
 
 
-def _hanning(da, N):
+def _hanning(da, N, dim):
     """Apply Hanning window"""
 
-    window = np.hanning(N[-1]) * np.hanning(N[-2])[:, np.newaxis]
+    if len(N)==1:
+        window = xr.DataArray(np.hanning(N[0]), dims=dim)
+    elif len(N)==2:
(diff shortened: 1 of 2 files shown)


rabernat:
Please rename this pull request to describe what it does. Does it fix a specific issue? If so, you can refer to the issue in the description (e.g. #16).

xrft/xrft.py
```diff
@@ -13,10 +13,32 @@
...
+def _hanning(da, N, dim):
     """Apply Hanning window"""
 
-    window = np.hanning(N[-1]) * np.hanning(N[-2])[:, np.newaxis]
+    if len(N)==1:
+        window = xr.DataArray(np.hanning(N[0]), dims=dim)
+    elif len(N)==2:
+        window = xr.DataArray(np.hanning(N[-1])
+                            * np.hanning(N[-2])[:,np.newaxis],
+                            dims=dim)
```
rabernat: This if / else syntax can probably be avoided by using list comprehensions and reduce.

roxyboy: I'm not sure if that's possible as the broadcasting is different with data that have different dimensions.

roxyboy:
I added tests for the _hanning function.

rabernat:
#21 implements a different fix for the same issue.

roxyboy:
Your new PR for windowing is better so I'm going to close this pull request.

PR #20, opened 2017-07-18
fix dask and coordinate issues

### LINKED ISSUE #17: Problems with dimensions without coordinates
Hey, 
I just ran into a problem with an array that has dimensions w/o coordinates. If I run this little example:
```
import xrft
import xarray as xr
import numpy as np

data = xr.DataArray(np.random.random([20,30,100]),dims=['time','lat','lon'])
ps = xrft.power_spectrum(data,dim=['time'])
```

It gives a key error:
```
KeyErrorTraceback (most recent call last)
<ipython-input-17-236267e01ceb> in <module>()
      6 data = xr.DataArray(np.random.random([20,30,100]),dims=['time','lat','lon'])
      7 print data
----> 8 ps = xrft.power_spectrum(data,dim=['time'])
      9 
     10 # # But that is not the main issue here

/home/julius/code/conda_envs/src/xrft/xrft/xrft.pyc in power_spectrum(da, dim, shift, detrend, density, window)
    273     daft = dft(da,
    274             dim=dim, shift=shift, detrend=detrend,
--> 275             window=window)
    276 
    277     coord = list(daft.coords)

/home/julius/code/conda_envs/src/xrft/xrft/xrft.pyc in dft(da, dim, shift, detrend, window)
    219             newcoords[d] = da.coords[d].values
    220         else:
--> 221             newcoords[d] = k_coords[d]
    222 
    223     dk = [l[1] - l[0] for l in k]

KeyError: 'lat'
```

If I run the same example with coordinates it works fine:

```
import xrft
import xarray as xr
import numpy as np

data = xr.DataArray(np.random.random([20,30,100]),dims=['time','lat','lon'],coords={'time':range(20),'lat':range(30),'lon':range(100)})
ps = xrft.power_spectrum(data,dim=['time'])

```

Best Julius

### LINKED ISSUE #19: dask laziness is broken!
A key behavior of xrft is that it is lazy: it does not convert dask data to numpy data until forced to.

This behavior appears to be broken in the latest version:

```python
Nx = 16
Lx = 1.0
x = np.linspace(0, Lx, Nx)
coords = [x]
da = xr.DataArray(np.random.rand(Nx), coords=coords, dims=['x']).chunk()
dft = xrft.dft(da)
print(type(da.data))
print(type(dft.data))
```

gives
```
<class 'dask.array.core.Array'>
<type 'numpy.ndarray'>
```

rabernat:
fixes #17 
fixes #19

This PR adds add some additional checks for dask laziness and also handles situations without coordinate variables.

It is the beginning of a more thorough refactoring of the tests.

### DIFF
--- a/xrft/xrft.py
+++ b/xrft/xrft.py
@@ -12,6 +12,11 @@ __all__ = ["detrend2","_detrend_wrap",
             "isotropic_powerspectrum","isotropic_crossspectrum",
             "fit_loglog"]
 
+def _fft_module(da):
+    if da.chunks:
+        return dsar.fft
+    else:
+        return np.fft
 
 def _hanning(da, N):
     """Apply Hanning window"""
@@ -146,8 +151,12 @@ def dft(da, dim=None, shift=True, detrend=None, window=False):
     daft : `xarray.DataArray`
         The output of the Fourier transformation, with appropriate dimensions.
     """
-    if np.isnan(da.values).any():
-        raise ValueError("Data cannot take Nans")
+    # we can't do da.values because it
+    if not da.chunks:
+        if np.isnan(da.values).any():
+            raise ValueError("Data cannot take Nans")
+
+    fft = _fft_module(da)
 
     if dim is None:
         dim = da.dims
@@ -171,6 +180,7 @@ def dft(da, dim=None, shift=True, detrend=None, window=False):
                              "coodinate %s is not evenly spaced" % d)
         delta_x.append(delta)
     # calculate frequencies from coordinates
+    # coordinates are always loaded eagerly, so we use numpy
     k = [ np.fft.fftfreq(Nx, dx) for (Nx, dx) in zip(N, delta_x) ]
 
     if detrend == 'constant':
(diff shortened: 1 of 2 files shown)


xrft/xrft.py
```diff
@@ -171,14 +180,15 @@ def dft(da, dim=None, shift=True, detrend=None, window=False):
...
     # calculate frequencies from coordinates
+    # coordinates are always loaded eagerly, so we use numpy
     k = [ np.fft.fftfreq(Nx, dx) for (Nx, dx) in zip(N, delta_x) ]
 
     if detrend == 'constant':
         da = da - da.mean(dim=dim)
     elif detrend == 'linear':
         if hasattr(da.data, 'dask'):
             func = _detrend_wrap(detrend2)
-            da = xr.DataArray(func(da.data, axes=axis_num).compute(),
```
rabernat: @roxyboy: this line was breaking the laziness. If you call compute on the dask array, there is no point in using dask!

roxyboy: Do I just keep the same code with removing `compute` then??

rabernat:
No comments so I will merge.

PR #21, opened 2017-07-18
Fix windowing

### LINKED ISSUE #16: Problem with windowing

rabernat:
fixes #16 

This PR introduces a new approach to creating windows via xarray dataarrays, which automatically handle broadcasting in arbitrary dimensions.

I added a simple test of @jbusecke's issue (#16). But more comprehensive tests of windowing, using different combinations of dimensions, should probably be added.

### DIFF
--- a/xrft/xrft.py
+++ b/xrft/xrft.py
@@ -3,9 +3,12 @@ import xarray as xr
 import pandas as pd
 import functools as ft
 import dask.array as dsar
+from dask import delayed
 import scipy.signal as sps
 import scipy.linalg as spl
 import warnings
+from functools import reduce
+import operator
 
 __all__ = ["detrend2","_detrend_wrap",
             "dft","power_spectrum","cross_spectrum",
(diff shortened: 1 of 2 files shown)


roxyboy:
I've never used the `dask.delayed` feature but looks like a cool PR

xrft/xrft.py
```diff
@@ -198,7 +196,8 @@ def dft(da, dim=None, shift=True, detrend=None, window=False):
                 raise ValueError("Data should be dask array.")
 
     if window:
-        da = _hanning(da, N)
+        window_array = _create_window(da, dim)
+        da *= window_array
```
roxyboy: One thing I've noticed previously when windowing was that it was safer to have `(new variable name) = da*window_array` instead of `da *= window_array` for some reason... If add more tests, we might find out that this doesn't matter anymore for your new windowing function

rabernat: This is a great suggestion. I will fix it.

PR #22, opened 2017-07-21
Three-dimensional detrending

roxyboy:
I added the option to apply detrending over 3 dimensions simultaneously, instead of just a 2D plane.

### DIFF
--- a/xrft/xrft.py
+++ b/xrft/xrft.py
@@ -10,7 +10,7 @@ import warnings
 from functools import reduce
 import operator
 
-__all__ = ["_detrend","_detrend_wrap",
+__all__ = ["detrendn","detrend_wrap",
             "dft","power_spectrum","cross_spectrum",
             "isotropic_powerspectrum","isotropic_crossspectrum",
             "fit_loglog"]
@@ -43,7 +43,7 @@ def _apply_window(da, dims, window_type='hanning'):
 
     return da * reduce(operator.mul, windows[::-1])
 
-def _detrend(da, axes=None):
+def detrendn(da, axes=None):
     """
     Detrend by subtracting out the least-square plane or least-square cubic fit
     depending on the number of axis.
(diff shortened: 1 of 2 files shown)


roxyboy:
I also refactored to windowing function a bit.

xrft/xrft.py
```diff
@@ -196,12 +196,9 @@ def dft(da, dim=None, shift=True, detrend=None, window=False):
                 raise ValueError("Data should be dask array.")
 
     if window:
-        window_array = _create_window(da, dim)
-        da_win = da * window_array
-    else:
-        da_win = da
+        da = _apply_window(da, dim)
```
roxyboy: I refactored the windowing function a bit.

xrft/xrft.py
```diff
@@ -10,7 +10,7 @@
 from functools import reduce
 import operator
 
-__all__ = ["detrend2","_detrend_wrap",
+__all__ = ["_detrend","_detrend_wrap",
```
rabernat: the underscore indicates a private function. Such functions shouldn't be part of the public API

xrft/xrft.py
```diff
@@ -97,25 +122,42 @@ def func(a, axes=None):
...
+def _apply_detrend(da, axis_num):
+    """Wrapper function for applying detrending"""
+    if da.chunks:
+        func = _detrend_wrap(_detrend)
+        da = xr.DataArray(func(da.data, axes=axis_num),
+                        dims=da.dims, coords=da.coords)
+    else:
+        if da.ndim == 1:
+            da = xr.DataArray(sps.detrend(da),
+                            dims=da.dims, coords=da.coords)
```
rabernat: So you use scipy for 1D detrending and your own custom function for higher dimensions?

roxyboy: I ended up doing that because the scipy function doesn't take data with more than one dimension.

PR #25, opened 2017-10-30
Added tolerance option to powerspectrum functions to give option to

kaipak:
Added parameter to power spectrum functions so you can control tolerance on even spacing requirement for Fourier transform.

### DIFF
--- a/.cache/v/cache/lastfailed
+++ b/.cache/v/cache/lastfailed
@@ -1,0 +1,4 @@
+{
+  "xrft/tests/test_xrft.py::test_stol": true,
+  "xrft/tests/test_xrft.py::test_stol[dask]": true
+}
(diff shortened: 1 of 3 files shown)


rabernat:
Fantastic, thanks! However, I'm not sure why this did not decrease test coverage...

Could you add a quick test of this new parameter?

kaipak:
For sure! I'll have to figure out how to do that :).

rabernat:
Just write a new function at the end of `tests/test_xrft.py` called something like `test_stol` which checks that the parameter behaves as expected (i.e. raises  or doesn't raise errors depending on value of stol and coordinate spacing). 

The spacing issue is already addressed in this test:
https://github.com/rabernat/xrft/blob/master/xrft/tests/test_xrft.py#L149

You probably want to remove that and put something similar into the new standalone function.

PR #26, opened 2017-11-03
Add spacing_tol parameter to dft

kaipak:
As requested, added some tests for new parameter option.

### DIFF
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -586,7 +586,6 @@ def test_spacing_tol(test_data_1d):
     da = test_data_1d
     da2 = da.copy().load()
 
-    # test for changing tolerance levels
     # Create improperly spaced data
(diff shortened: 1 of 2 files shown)


xrft/tests/test_xrft.py
```diff
@@ -587,3 +581,34 @@ def test_isotropic_cs():
...
+    # This shouldn't raise an error
+    xrft.dft(da3, spacing_tol=1e-1)
+    # But this should 
+    with pytest.raises(ValueError):
+        xrft.dft(da3, spacing_tol=1e-4)
+
+    # modify data to be non-evenly spaced
+    #da2[-1] = np.nan
+    #with pytest.raises(ValueError):
+    #    ft = xrft.dft(da2)
```
rabernat: let's just clean up these comments...we have a record of the code history in git

Issue #49, opened 2018-08-28
operating on chunked data in segments


rabernat:
Imagine if you are trying to do the power spectrum of a really long timeseries. So long that you don't want to put it all in memory. Currently xgcm / dask fail when you try to take an fft over a chunked dimension:

```python
import numpy as np
import dask.array as dsa
import xarray as xr
import xrft
da = xr.DataArray(np.random.rand(1000), dims=['x']).chunk({'x': 100})
xrft.power_spectrum(da, dim='x')
```
```
ValueError: Dask array only supports taking an FFT along an axis that 
has a single chunk. An FFT operation was tried on axis 0 
which has chunks (100, 100, 100, 100, 100, 100, 100, 100, 100, 100). To change the array's chunks use dask.Array.rechunk.
```

This makes sense if you care about the full Fourier transform. But often with long timeseries, we want to do some sort of tapering in which we average over multiple segments in order to get a less biased estimate of the power spectrum (like in [Bartlett's method](https://en.wikipedia.org/wiki/Bartlett%27s_method)). In this case, the chunks provide a natural way to split up the full interval.

With the example above, we can do the following:
```python
# get raw data
data = da.data
# reshape so there is one chunk for each item on axis 0
data_rs = data.reshape((10, 100))
# transform along the last axis
dsa.fft.fft(data_rs, axis=1)
```
this works and returns
```
dask.array<fft, shape=(10, 100), dtype=complex128, chunksize=(1, 100)>
```
You could then apply Bartlett's method by averaging over axis 0.

In xrft, we could try to accommodate this by adding a new keyword
```python
xrft.power_spectrum(da, dim='x', chunks_to_segments=True)
```
which would then return something like
```
<xarray.DataArray '...' (x_segment: 10, freq_x: 100)>
dask.array<shape=(10, 100), dtype=complex128, chunksize=(1, 100)>
Coordinates:
  * freq_x          (freq_x) float64 -0.5 -0.499 -0.498 -0.497 -0.496 -0.495 ...
Dimensions without coordinates: x_segment
```

I think this would be highly useful!

PR #51, opened 2018-08-28
dft for chunked data along the axis of FT

roxyboy:
@rabernat Something like this for Issue #49 ?


xrft/xrft.py
```diff
@@ -177,17 +178,22 @@ def dft(da, spacing_tol=1e-3, dim=None, shift=True, detrend=None, window=False):
...
-    window : bool (optional)
+    window : bool (default)
         Whether to apply a Hann window to the data before the Fourier
         transform is taken. A window will be applied to all the dimensions in
         dim.
+    chunks_to_segments : bool (default)
+        Whether the data is chunked along the axis to take FFT.
+    kwargs : dict
+        Need to provide the information of chunk length (`seglen`: int)
+        when `chunks_to_segment=True`.
```
rabernat: a) we shouldn't hide arguments in `kwargs` unnecessarily

b) we shouldn't need to specify `seglen` at all! It should be determined by dask chunks themselves.

xrft/xrft.py
```diff
@@ -209,6 +215,33 @@ def dft(da, spacing_tol=1e-3, dim=None, shift=True, detrend=None, window=False):
         dim = da.dims
 
     # the axes along which to take ffts
+    if chunks_to_segments:
+        data = da.data
+        if len(dim)>1:
+            raise NotImplementedError("Currently, only a one-dimensional "
+                                     "DFT is supported when the dimension "
+                                     "to FT is chunked.")
```
rabernat: I don't see why we need this limitation. It should be possible to to 2d dft on 2d chunks (or any number of dimensions).

One needs to check each dimension in `dim` to see if reshaping is needed.

xrft/xrft.py
```diff
@@ -209,6 +215,33 @@ def dft(da, spacing_tol=1e-3, dim=None, shift=True, detrend=None, window=False):
...
+        suffix = '_segment'
+        newdims = []
+        newcoords = {}
+        newshape = []
+        for d in da.dims:
+            if d == dim[0]:
+                newdims.append(d + suffix)
+                newdims.append(d)
+                n = len(da[d])
+                seglen = kwargs['seglen']
```
rabernat: Here we should instead figure out the chunks by examining the dask chunks.

An error should be raised if the chunks are not uniform.

xrft/xrft.py
```diff
@@ -209,6 +215,33 @@ def dft(da, spacing_tol=1e-3, dim=None, shift=True, detrend=None, window=False):
...
+                coord_rs = da[d].data.reshape((int(n/seglen),int(seglen)))
+                newshape.append(int(n/seglen))
+                newshape.append(int(seglen))
+                newcoords[d+suffix] = range(int(n/seglen))
+                newcoords[d] = coord_rs[0]
+            else:
+                newdims.append(d)
+                newshape.append(len(da[d]))
+                newcoords[d] = da[d].data
+        da = xr.DataArray(data.reshape(newshape), dims=newdims, coords=newcoords)
```
rabernat: I recommend putting the logic in the above ~20 lines into a standalone private function

```
def _stack_chunks(da, dim, suffix='_segment):
    """Reshape a DataArray so there is only one chunk along dimension `dim`
    ...
    """
```

xrft/xrft.py
```diff
@@ -309,6 +342,11 @@ def power_spectrum(da, spacing_tol=1e-3, dim=None, shift=True, detrend=None, den
     window : bool (optional)
         Whether to apply a Hann window to the data before the Fourier
         transform is taken
+    chunks_to_segments : bool (default)
+        Whether the data is chunked along the axis to take FFT.
+    kwargs : dict
+        Need to provide the information of chunk length (`seglen`: int)
+        when `chunks_to_segment=True`.
```
rabernat: Don't use `kwargs` this way.

You should review [the zen of python](https://www.python.org/dev/peps/pep-0020/). Line 2: "Explicit is better than implicit."

xrft/tests/test_xrft.py
```diff
@@ -196,6 +197,14 @@ def test_dft_4d():
     with pytest.raises(NotImplementedError):
         xrft.dft(da, dim=['time','y','x'], detrend='linear')
 
+    with pytest.raises(NotImplementedError):
+        xrft.dft(da.chunk({'time':8}), dim=['time','z'], chunks_to_segments=True)
+    ft = xrft.dft(da.chunk({'time':8}), dim=['time'], shift=False,
+                 chunks_to_segments=True, seglen=8)
+    data = da.chunk({'time':8}).data.reshape((2,8,16,16,16))
+    npt.assert_almost_equal(ft.values, dsar.fft.fftn(data, axes=[1]).compute(),
+                           decimal=7)
```
rabernat: Rather than absorbing this into an existing test, I would prefer to have a *new* test function for this feature.

xrft/xrft.py
roxyboy: The current major issue with this is that the 2D, 3D detrending won't work if the axes are chunked. I'll see if I can change this easily.

roxyboy:
Nevermind, got linear detrending for 2D cases to work.

rabernat:
As far as I understand it, you should first transform the data to have reshaped dimensions, and then all the old windowing / detrending stuff should just work. It's really a pre-processing step.

xrft/tests/test_xrft.py
```diff
@@ -98,17 +99,17 @@ def test_detrend():
...
+    # d5d = d4d[np.newaxis,:,:,:,:] + s[:,np.newaxis,np.newaxis,
+    #                                     np.newaxis,np.newaxis]
+    # da5d = xr.DataArray(d5d, dims=['s','time','z','y','x'],
+    #                  coords={'s':range(len(s)),'time':range(len(t)),
+    #                          'z':range(len(z)),'y':range(len(y)),
+    #                          'x':range(len(x))}
+    #                    )
+    # da = da5d.chunk({'time':1})
+    # with pytest.raises(ValueError):
+    #     func(da.data).compute()
```
rabernat: Why is this commented out?

xrft/tests/test_xrft.py
```diff
@@ -184,17 +185,17 @@ def test_dft_4d():
...
     npt.assert_almost_equal(ft.values, np.fft.fftn(da.values))
 
-    with pytest.raises(NotImplementedError):
-        xrft.dft(da, detrend='linear')
-    with pytest.raises(NotImplementedError):
-        xrft.dft(da, dim=['time','y','x'], detrend='linear')
+    # with pytest.raises(NotImplementedError):
+    #     xrft.dft(da, detrend='linear')
+    # with pytest.raises(NotImplementedError):
+    #     xrft.dft(da, dim=['time','y','x'], detrend='linear')
```
rabernat: why do we have commented-out code here?

xrft/tests/test_xrft.py
```diff
@@ -203,6 +204,31 @@ def test_dft_4d():
...
+def test_bartlett():
+    N = 32
+    da = xr.DataArray(np.random.rand(N,N,N),
+                     dims=['time','y','x'],
+                     coords={'time':range(N),'y':range(N),'x':range(N)}
+                     )
+
+    # with pytest.raises(NotImplementedError):
+    #     xrft.dft(da.chunk({'time':16}), dim=['time','y'], detrend='linear',
+    #             chunks_to_segments=True)
```
rabernat: ditto

roxyboy:
Yea, so everything is working fine now will all your reviews incorporated :)

xrft/tests/test_xrft.py
```diff
@@ -203,6 +204,31 @@ def test_dft_4d():
...
+    ft = xrft.dft(da.chunk({'time':16}), dim=['time'], shift=False,
+                 chunks_to_segments=True)
+    data = da.chunk({'time':16}).data.reshape((2,16,N,N))
+    npt.assert_almost_equal(ft.values, dsar.fft.fftn(data, axes=[1]),
+                           decimal=7)
+    ft = xrft.dft(da.chunk({'y':16,'x':16}), dim=['y','x'], shift=False,
+                 chunks_to_segments=True)
+    data = da.chunk({'y':16,'x':16}).data.reshape((N,2,16,2,16))
+    npt.assert_almost_equal(ft.values, dsar.fft.fftn(data, axes=[2,4]),
+                           decimal=7)
```
rabernat: I would also check that the outputs have the expected dimension names, sizes, and orders.

xrft/xrft.py
```diff
@@ -112,9 +112,8 @@ def detrend_wrap(detrend_func):
     Wrapper function for `xrft.detrendn`.
     """
     def func(a, axes=None):
-        if a.ndim > 4 or len(axes) > 3:
-            raise ValueError("Data has too many dimensions "
-                            "and/or too many axes to detrend over.")
+        if len(axes) > 3:
+            raise ValueError("Data has too many axes to detrend over.")
```
rabernat: Let's be more explicit with this error, e.g. "Detrending is only possible over 3 dimensions or less, but input had 5 dimensions."

xrft/xrft.py
```diff
@@ -159,7 +158,39 @@ def _apply_detrend(da, axis_num):
...
     return da
 
-def dft(da, spacing_tol=1e-3, dim=None, shift=True, detrend=None, window=False):
+def _stack_chunks(da, dim, suffix='_segment'):
+    """Reshape a DataArray so there is only one chunk along dimension `dim`"""
+    data = da.data
+    # if len(dim)>1 and detrend=='linear':
+    #     raise NotImplementedError("Currently, only a one-dimensional "
+    #                              "linear detrending is supported "
+    #                              "when the dimension to FT is chunked.")
```
rabernat: get rid of comments

xrft/xrft.py
```diff
@@ -159,7 +158,39 @@ def _apply_detrend(da, axis_num):
...
+    newdims = []
+    newcoords = {}
+    newshape = []
+    for d in da.dims:
+        if d in dim:
+            axis_num = da.get_axis_num(d)
+            if np.diff(da.chunks[axis_num]).sum() != 0:
+                raise ValueError("Chunk lengths need to be the same.")
```
rabernat: More verbose: "Chunk lengths must be the same for `chunks_to_segments` to work, but instead got chunks (X, Y, Z) along dimension 'time'")

xrft/xrft.py
```diff
@@ -159,7 +158,39 @@ def _apply_detrend(da, axis_num):
...
+            n = len(da[d])
+            chunklen = da.chunks[axis_num][0]
+            coord_rs = da[d].data.reshape((int(n/chunklen),int(chunklen)))
+            newdims.append(d + suffix)
+            newdims.append(d)
+            newshape.append(int(n/chunklen))
+            newshape.append(int(chunklen))
+            newcoords[d+suffix] = range(int(n/chunklen))
```
rabernat: I don't think you need explicit coords here. Xarray now accepts dimensions without coordinates.

xrft/xrft.py
```diff
@@ -159,7 +158,39 @@ def _apply_detrend(da, axis_num):
...
+            newcoords[d] = coord_rs[0]
+        else:
+            newdims.append(d)
+            newshape.append(len(da[d]))
+            newcoords[d] = da[d].data
+
+    return xr.DataArray(data.reshape(newshape), dims=newdims, coords=newcoords)
```
rabernat: Really nice function! 👍

rabernat: What happens to attributes that may have been in the original array? Are they lost? I would try to preserve the attributes.

xrft/xrft.py
```diff
@@ -234,6 +270,9 @@ def dft(da, spacing_tol=1e-3, dim=None, shift=True, detrend=None, window=False):
     if detrend == 'constant':
         da = da - da.mean(dim=dim)
     elif detrend == 'linear':
+        for d in da.dims:
+            if d not in dim:
+                da = da.chunk({d:1})
```
rabernat: Can you help me understand why you need to do this?

xrft/xrft.py
rabernat: Still need to change this doc string to match the function signature.

xrft/xrft.py
```diff
@@ -324,9 +368,14 @@ def power_spectrum(da, spacing_tol=1e-3, dim=None, shift=True, detrend=None, den
...
-            dim=dim, shift=shift, detrend=detrend,
-            window=window)
+    if chunks_to_segments:
+        daft = dft(da, spacing_tol,
+                  dim=dim, shift=shift, detrend=detrend, window=window,
+                  chunks_to_segments=chunks_to_segments)
+    else:
+        daft = dft(da, spacing_tol,
+                  dim=dim, shift=shift, detrend=detrend,
+                  window=window)
```
rabernat: Why do you need a special case? Can't you just pass `chunks_to_segments=chunks_to_segments`?

rabernat:
I guess we probably eventually want to add the `chunks_to_segments` options to `cross_spectrum` and `isotropic_power_spectrum` too, no?

xrft/tests/test_xrft.py
roxyboy: I'll just delete this as it is implemented.

xrft/tests/test_xrft.py
roxyboy: Also implemented so I'll just delete it.

xrft/tests/test_xrft.py
roxyboy: The current code should be able to take data that has more than 5 dimensions now so I commented it out.

roxyboy:
Added `chunks_to_segments` to `isotropic_power_spectrum` would fail currently due to the same reasonings as issue #9 . I haven't been able to parallelize the algorithm for azimuthal averaging.

xrft/xrft.py
roxyboy: The linear algebra for detrending currently can only take one chunk at a time.

xrft/xrft.py
roxyboy: It works now with more than data that has 4 dimensions. Detrending is only supported up to 3 dimensions at the moment.

xrft/xrft.py
rabernat: But why does dask have to be involved? Are we using map_blocks to apply the detrending?

I'm worried about this. It is generally not a good idea to silently rechunk the user's data. What if this accidentally creates a billion chunks. I would prefer to raise an error and make the user do it.

roxyboy: Yes, we use map_blocks.

Ok, I'll make it raise an error but modify it so that it'll automatically chunk the `_segment` dimensions.

xrft/xrft.py
```diff
@@ -232,20 +237,27 @@ def dft(da, spacing_tol=1e-3, dim=None, shift=True, detrend=None, window=False,
...
         if np.isnan(da.values).any():
             raise ValueError("Data cannot take Nans")
+    else:
+        if detrend=='linear' and len(dim)>1:
+            for d in da.dims:
+                a_n = da.get_axis_num(d)
+                if d not in dim and da.chunks[a_n][0]>1:
+                    raise ValueError("Linear detrending utilizes the `dask.map_blocks` "
+                                    "API so the dimensions not being detrended "
+                                    "must have the chunk length of 1.")
```
roxyboy: @rabernat I added this to raise an error if the user didn't provide data chunked with lengths of one for dimensions that won't be detrended.

xrft/xrft.py
```diff
@@ -283,17 +273,10 @@ def dft(da, spacing_tol=1e-3, dim=None, shift=True, detrend=None, window=False,
     if detrend == 'constant':
         da = da - da.mean(dim=dim)
     elif detrend == 'linear':
+        for d in da.dims:
+            if d not in dim:
+                da = da.chunk({d:1})
```
roxyboy: Although this may be a bit redundant, I kept it to make sure the chunks have unit length.

rabernat: For the future, I recommend we use [xarray.apply_ufunc](http://xarray.pydata.org/en/stable/generated/xarray.apply_ufunc.html) to apply detrending, rather than map blocks. But that can wait for a future PR.

xrft/tests/test_xrft.py
```diff
@@ -184,25 +173,52 @@ def test_dft_4d():
...
-        xrft.dft(da, dim=['time','y','x'], detrend='linear')
-
     da_prime = xrft.detrendn(da[:,0].values, [0,1,2]) # cubic detrend over time, y, and x
     npt.assert_almost_equal(xrft.dft(da[:,0].drop('z'),
                                     dim=['time','y','x'],
                                     shift=False, detrend='linear'
                                     ).values,
                             np.fft.fftn(da_prime))
 
+def test_bartlett():
```
rabernat: can we just give this test a more informative name. Like `test_chunks_to_segments()`?

xrft/xrft.py
roxyboy: so... we can merge this?

xrft/xrft.py
```diff
@@ -177,17 +204,19 @@ def dft(da, spacing_tol=1e-3, dim=None, shift=True, detrend=None, window=False):
...
-    shift : bool (optional)
+    shift : bool (default)
         Whether to shift the fft output.
     detrend : str (optional)
         If `constant`, the mean across the transform dimensions will be
         subtracted before calculating the Fourier transform (FT).
         If `linear`, the linear least-square fit will be subtracted before
         the FT.
-    window : bool (optional)
+    window : bool (default)
```
rabernat: why did you change these from `optional` to `default`? That doesn't match the [numpydoc spec](https://numpydoc.readthedocs.io/en/latest/format.html#sections). In fact, looking closely at the spec, it should be
```
window: bool, optional
```

xrft/tests/test_xrft.py
```diff
@@ -184,25 +173,52 @@ def test_dft_4d():
...
+    data = da.chunk({'y':16,'x':16}).data.reshape((N,2,16,2,16))
+    npt.assert_almost_equal(ft.values, dsar.fft.fftn(data, axes=[2,4]),
+                           decimal=7)
+    ps = xrft.power_spectrum(da.chunk({'y':16,'x':16}), dim=['y','x'],
+                            shift=False, density=False,
+                            chunks_to_segments=True)
+    npt.assert_almost_equal(ps.values,
+                           (ft*np.conj(ft)).real.values,
+                           )
+
```
rabernat: There is no test for `cross_spectrum`.

roxyboy:
Ok, the tests are passing so I'm gonna merge this.

PR #54, opened 2019-02-26
Option to use real fft

TomNicholas:
Added an option to do a real fft instead of assuming that you always wish to retain negative frequencies.

Required wrapping `np.fft.rfftn` instead of `np.fft.fftn`, and setting `shift=False` if `real=True`.

- [x] Tests (for 1d data)
- [x] Docstring

### DIFF
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -101,6 +101,9 @@ def test_detrend():
 
 def test_dft_1d(test_data_1d):
     """Test the discrete Fourier transform function on one-dimensional data."""
+
+    # TODO group these tests into a parameterized class
+
     da = test_data_1d
     Nx = len(da)
     dx = float(da.x[1] - da.x[0]) if 'x' in da.dims else 1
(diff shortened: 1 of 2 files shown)


rabernat:
Wow @TomNicholas, thanks a lot for this PR! It looks very complete, well tested and documented.

I have not yet gone through your code in high detail, but just based on the docstring for the `real` keyword, I do have a concern about multidimensional real FTs. In higher dimensions, the symmetry is not so simple that you can just discard the negative frequencies. [These notes](https://www2.ph.ed.ac.uk/~wjh/teaching/Fourier/documents/symmetry.pdf) provide a good discussion of the issue.

Have you thought about how this new feature behaves in higher dimensions?

roxyboy:
Yea, dropping the negative frequencies for higher dimensions may become an issue.

TomNicholas:
> Have you thought about how this new feature behaves in higher dimensions?

@rabernat I had never actually realised that about Fourier transforms, but I think I've solved it!

My docstring implies that I'm manually dropping all negative frequencies but I'm actually not, I'm just returning the result of `np.fft.rfftn`. This function actually uses `np.fft.fft` on all axes but the last, on which it uses `np.fft.rfft`. That means in dimensions > 1 it will retain some negative frequencies rather than dropping everything except the positive quadrant.

So I changed the calculation of the returned frequency coordinates to use `np.fft.fftfreq` for all but the last dimension, where it uses `np.fft.rfftfreq`. This will effectively drop only the frequencies which are consistent with symmetry in multiple dimensions.

I've implemented all of this, changed the docstring and added a test for the 2d case to show that it works.

PR #65, opened 2019-04-01
switch to versioneer

roxyboy:
I changed the version control to versioneer as suggested by #65.


rabernat:
I'm happy to see you working on this.

If you'd like feedback from the other devs, you might consider
- adding descriptions to your PRs
- waiting at least a day before merging

roxyboy:
The code coverage decreased due to `_version.py`. It comes from the versioneer package but is that something I should also have tests for @rabernat ?

rabernat:
Consider excluding it from coverage, as done here:

https://github.com/xgcm/xgcm/blob/master/.coveragerc

setup.py
```diff
@@ -1,29 +1,74 @@
 import os
+import versioneer
 from setuptools import setup, find_packages
 PACKAGES = find_packages()
 
-# Get version and release info, which is all stored in xrft/version.py
-ver_file = os.path.join('xrft', 'version.py')
-with open(ver_file) as f:
-    exec(f.read())
+VERSION = 'v0.1'
```
rabernat: Might want to remove this `VERSION` variable, since it is now unused.

setup.py
```diff
@@ -1,29 +1,74 @@
...
+#             author_email=AUTHOR_EMAIL,
+#             platforms=PLATFORMS,
+#             version=VERSION,
+#             packages=PACKAGES,
+#             package_data=PACKAGE_DATA,
+#             requires=REQUIRES)
+#
+#
+# if __name__ == '__main__':
+#     setup(**opts)
```
rabernat: Remove all these comments please.

rabernat:
What version is versioneer giving you now that this has been implemented?

```
import xrft
print(xrft.__version__)
```

roxyboy:
It gives `0.1-beta+76.g9bbebbc.dirty`

setup.py
roxyboy: Both changes are implemented.

Issue #57, opened 2019-03-26
Support real='coord'


dcherian:
Currently the `real` kwarg is boolean, and takes a real FFT along the last dimension. This means that the user needs to transpose the data to an appropriate shape. 

I propose that xrft support the following syntax:

`xrft.xrft.dft(da, dim=['time', 'longitude'], real='time')`

roxyboy:
Thanks for pointing that out @dcherian. As of now, the `real` kwargs implements the `numpy.rfftn` function in paralyzation which only takes the real FFT along the last dimension: https://docs.scipy.org/doc/numpy/reference/generated/numpy.fft.rfftn.html#numpy.fft.rfftn

dcherian:
I should've been clearer. 
`xrft.xrft.dft(da, dim=['time', 'longitude'], real='time')` should be equivalent to
`xrft.xrft.dft(da.transpose(['longitude', 'time']), dim=['time', 'longitude'], real=True)`

i.e. xrft does the appropriate transpose based on the `real` kwarg.

roxyboy:
That makes sense for two dimensional arrays but for higher dimensional arrays, I'm not sure if the transposing should be made automatically...

We could either implement a warning when the `real` kwarg is used or automate the transposition.

rabernat:
@dcherian, we really appreciate your feedback on this. There are not many people working on xrft...just me and @roxyboy (and I don't have much time for it these days).

We would be very happy to get a PR from you that addresses your issue. 😄

rabernat:
Actually, @TomNicholas is the one who added the `real` option in #54, so maybe he would be a good person to address this.

TomNicholas:
> I propose that xrft support the following syntax:
>
> `xrft.xrft.dft(da, dim=['time', 'longitude'], real='time')`

I think that API would be better than what I went for! Is there any performance/consistency reason why we shouldn't just transpose then apply `np.fft.rfftn`? If we're happy to transpose then this seems like a simple improvement to me?

> There are not many people working on xrft...just me and roxyboy (and I don't have much time for it these days).

At the risk of totally derailing the discussion.... Fourier transforms (and the xhistogram module I saw you started @rabernat) are so clearly generally-useful that I think we should push for these kind of tools to be packaged more centrally for the convenience of xarray's userbase. Not as part of xarray itself but more like the new scipy to xarray's numpy. Otherwise we're just going to end up with lots of xarray users reinventing the wheel over and over (which I think is already happening).

[...]

PR #66, opened 2019-04-05
implementing real=coords

roxyboy:
I implemented #57 and the tests are passing. I made the `real` argument so that the user gives a list of the order how they would want the input transposed and returns the real FFT with the same order of dimensions as the original input, e.g. the argument of `dim=['time'], real=['Y','X','time']` to take the real FFT over `time` for a dataset with dimensions of `['time','Y','X']` would return `['freq_time','Y','X']`.

I don't use real FFTs that much so would be great if @TomNicholas and @dcherian could try it out and see how it suits their needs :)

### DIFF
--- a/xrft/xrft.py
+++ b/xrft/xrft.py
@@ -209,10 +209,8 @@ def dft(da, spacing_tol=1e-3, dim=None, real=None, shift=True, detrend=None,
     dim : list, optional
         The dimensions along which to take the transformation. If `None`, all
         dimensions will be transformed.
-    real : list, optional
-        Default is `None` where no real Fourier transform (FT) will be taken.
-        Otherwise, reorder the dimensions to this order and uses the real FT
-        along the last dimesion.
+    real : str, optional
+        Real Fourier transform will be taken along this dimension.
     shift : bool, default
         Whether to shift the fft output. Default is `True`, unless `real=True`,
         in which case shift will be set to False always.
@@ -238,8 +236,15 @@ def dft(da, spacing_tol=1e-3, dim=None, real=None, shift=True, detrend=None,
         raise TypeError("Please provide a float argument")
 
     rawdims = da.dims
-    if real is not None and len(real)>0:
-        da = da.transpose(*real)
+    if real is not None:
+        transdim = list(rawdims)
+        if real not in transdim:
+            raise ValueError("The dimension along real FT is taken must "
+                            "be one of the existing dimensions.")
+        elif real != transdim[-1]:
+            transdim.remove(real)
+            transdim += [real]
+            da = da.transpose(*transdim)
     if dim is None:
         dim = da.dims
 
@@ -538,10 +543,10 @@ def cross_phase(da1, da2, spacing_tol=1e-3, dim=None, detrend=None,
                         'a single dimension.')
 
     daft1 = dft(da1, spacing_tol,
-                dim=dim, real=dim, shift=False, detrend=detrend,
+                dim=dim, real=dim[0], shift=False, detrend=detrend,
                 window=window, chunks_to_segments=chunks_to_segments)
     daft2 = dft(da2, spacing_tol,
-                dim=dim, real=dim, shift=False, detrend=detrend,
+                dim=dim, real=dim[0], shift=False, detrend=detrend,
                 window=window, chunks_to_segments=chunks_to_segments)
 
     if daft1.chunks and daft2.chunks:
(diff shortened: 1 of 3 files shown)


xrft/xrft.py
```diff
@@ -209,11 +209,12 @@ def dft(da, spacing_tol=1e-3, dim=None, real=False, shift=True, detrend=None,
     dim : list, optional
         The dimensions along which to take the transformation. If `None`, all
         dimensions will be transformed.
-    real : bool, optional
-        Whether the input array is all real or not. If set to True the
-        redundant negative frequencies will be discarded. Defaults to False.
+    real : list, optional
+        Default is `None` where no transposing will be applied. Otherwise,
```
TomNicholas: This docstring clearly explains the implementation, but users aren't concerned about the implementation, just the outcome. So you can maybe simplify to just saying something like "uses the real Fourier transform along this dimension"?

xrft/xrft.py
```diff
@@ -475,7 +489,7 @@ def cross_spectrum(da1, da2, spacing_tol=1e-3, dim=None,
     return cs
 
 
-def cross_phase(da1, da2, spacing_tol=1e-3, dim=None, detrend=None,
+def cross_phase(da1, da2, spacing_tol=1e-3, dim=None, real=None, detrend=None,
```
TomNicholas: Currently inside `cross_phase` then we always have `real=True`, so I don't think we need this extra parameter? 

In fact I think we should also change the `cross_phase` to only accept a dims argument of length one...

xrft/xrft.py
```diff
@@ -236,6 +237,9 @@ def dft(da, spacing_tol=1e-3, dim=None, real=False, shift=True, detrend=None,
     if not isinstance(spacing_tol, float):
         raise TypeError("Please provide a float argument")
 
+    rawdims = da.dims
+    if real is not None and len(real)>0:
+        da = da.transpose(*real)
```
TomNicholas: This is very neat I like it

xrft/xrft.py
```diff
@@ -331,7 +335,17 @@ def dft(da, spacing_tol=1e-3, dim=None, real=False, shift=True, detrend=None,
...
+    daft = xr.DataArray(f, dims=newdims, coords=newcoords)
+    if real is None:
+        return daft
+    else:
+        enddims = list(rawdims)
+        i = 0
+        for d in rawdims:
+            if d in dim:
+                enddims[i] = prefix + d
+            i += 1
```
TomNicholas: Couldn't this loop be replaced with a list comprehension? Something like:

```python
enddims = [prefix + d for d in rawdims if d in dim]
```

dcherian:
Hmmm... I don't see why the user needs to repeat the (possibly very long) list of dimensions. You could have real be a string kwarg instead as follows:

 ```
real = 'time'
dims = ['time', 'x', 'y']

if real and dims[-1] != real:
    dims.remove(real)
    dims += [real]
    da.transpose(*dims)
    transposed = True
else:
    transposed = False

# do fft
if real and transposed:
    # transpose back
```

**Comment by dcherian** [comment]
Do we want to ensure that `real is in dim`? i.e. the real axis  name needs to be specified twice once as `real` and once as `dim`. I.e. if `real not in dim`, it is not inserted into `dim`

File: xrft/xrft.py [hunk 5]
```diff
@@ -331,7 +344,12 @@ def dft(da, spacing_tol=1e-3, dim=None, real=False, shift=True, detrend=None,
...
-    return xr.DataArray(f, dims=newdims, coords=newcoords)
+    daft = xr.DataArray(f, dims=newdims, coords=newcoords)
+    if trans:
+        enddims = [prefix + d for d in rawdims if d in dim]
+        return daft.transpose(*enddims)
+    else:
+        return daft
 
 
 def power_spectrum(da, spacing_tol=1e-3, dim=None, shift=True, detrend=None, density=True,
```
dcherian: Add support for `real` here too?

xrft/xrft.py
```diff
@@ -236,8 +235,22 @@ def dft(da, spacing_tol=1e-3, dim=None, real=False, shift=True, detrend=None,
     if not isinstance(spacing_tol, float):
         raise TypeError("Please provide a float argument")
 
+    rawdims = da.dims
+    trans = False
+    if real is not None:
+        transdim = list(rawdims)
+        if real not in transdim:
+            raise ValueError("The dimension along real FT is taken must "
+                            "be one of the existing dimensions.")
```
dcherian: Print out names of existing dimensions in the error message?

roxyboy:
I added the `real` option for `power_spectrum` as suggested by @dcherian. Once the tests pass on Travis, I'm going to merge this.

PR #70, opened 2019-04-10
for loop for azimuthal averaging in isotropic functions

roxyboy:
I inserted a for loop for the `isotropic` functions so that it can take arrays up to 4 dimensions.

### DIFF
--- a/xrft/xrft.py
+++ b/xrft/xrft.py
@@ -511,7 +511,7 @@ def cross_phase(da1, da2, spacing_tol=1e-3, dim=None, detrend=None,
     .. math::
         da1' = da1 - \overline{da1};\ \ da2' = da2 - \overline{da2}
     .. math::
-        cp = \text{Arg} \mathbb{F}(da1')^* {\mathbb{F}(da2')}
+        cp = \text{Arg} [\mathbb{F}(da1')^*, \mathbb{F}(da2')]
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


rabernat:
I think we may be able to improve this in the future using my new [xhistogram](https://github.com/xgcm/xhistogram) package.

For now it's fine as is.

xrft/xrft.py
```diff
@@ -661,8 +658,25 @@ def isotropic_powerspectrum(da, spacing_tol=1e-3, dim=None, shift=True,
...
+        iso_ps = _azimuthal_avg(kidx, f, area, kr)
+    elif len(M) == 2:
+        for j in range(M[0]):
+            iso_ps[j] = _azimuthal_avg(kidx, f[j], area, kr)
+    elif len(M) == 3:
+        for j in range(M[0]):
+            for i in range(M[1]):
+                iso_ps[j,i] = _azimuthal_avg(kidx, f[j,i], area, kr)
+    else:
+        raise ValueError("Arrays with more than 4 dimensions is not supported.")
```
rabernat: _are_ not supported

xrft/xrft.py
```diff
@@ -742,8 +756,25 @@ def isotropic_crossspectrum(da1, da2, spacing_tol=1e-3,
...
+        iso_cs = _azimuthal_avg(kidx, f, area, kr)
+    elif len(M) == 2:
+        for j in range(M[0]):
+            iso_cs[j] = _azimuthal_avg(kidx, f[j], area, kr)
+    elif len(M) == 3:
+        for j in range(M[0]):
+            for i in range(M[1]):
+                iso_cs[j,i] = _azimuthal_avg(kidx, f[j,i], area, kr)
+    else:
+        raise ValueError("Arrays with more than 4 dimensions is not supported.")
```
rabernat: Since you have this code repeated in two functions, you should probably break it into a private standalone function.

PR #84, opened 2019-11-15
cleanup, keep coords

aaronspring:
- `dft` keeps coords, especially multi-dim coords
- remove `pytest-runner` because depreciated
- linting and rm unused

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,47 +1,51 @@
-import os
+from setuptools import find_packages, setup
+
 import versioneer
-from setuptools import setup, find_packages
+
 PACKAGES = find_packages()
 
-DISTNAME = 'xrft'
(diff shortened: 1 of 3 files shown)


aaronspring:
implemented based on suggestion from @rabernat

rabernat:
@aaronspring -- we really appreciate you work on this!

However, this PR is a bit hard to review because you have changed all our single quotes to double quotes. While certainly we welcome code-style improvements, I would prefer to address those in a standalone PR, rather than mixing them in with a new feature.

Could you please revert the style changes so we can focus on the new feature?

roxyboy:
The tests seem to fail because of the `cftime` package. Should we import this in the testing script?

Issue #82, opened 2019-11-10
xrft doesn't keep multidimensional coords


aaronspring:
Thanks for creating `xrft`.
I would like to use it on a dataset with multi-dim coords.

```python
ds = xr.tutorial.load_dataset('rasm')['Tair']
ds = ds.fillna(0) # Nans not allowed
xrft.power_spectrum(ds,dim='time')
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-48-9a6553143ee2> in <module>
----> 1 xrft.power_spectrum(ds,dim='time')

~/anaconda3/envs/xr/lib/python3.7/site-packages/xrft/xrft.py in power_spectrum(da, spacing_tol, dim, real, shift, detrend, window, chunks_to_segments, density)
    403     daft = dft(da, spacing_tol,
    404               dim=dim, real=real, shift=shift, detrend=detrend, window=window,
--> 405               chunks_to_segments=chunks_to_segments)
    406 
    407     if dim is None:

~/anaconda3/envs/xr/lib/python3.7/site-packages/xrft/xrft.py in dft(da, spacing_tol, dim, real, shift, detrend, window, chunks_to_segments)
    296             diff = diff.astype('timedelta64[s]').astype('f8')
    297         delta = diff[0]
--> 298         if not np.allclose(diff, diff[0], rtol=spacing_tol):
    299             raise ValueError("Can't take Fourier transform because "
    300                              "coodinate %s is not evenly spaced" % d)

~/anaconda3/envs/xr/lib/python3.7/site-packages/numpy/core/numeric.py in allclose(a, b, rtol, atol, equal_nan)
   2421 
   2422     """
-> 2423     res = all(isclose(a, b, rtol=rtol, atol=atol, equal_nan=equal_nan))
   2424     return bool(res)
   2425 

~/anaconda3/envs/xr/lib/python3.7/site-packages/numpy/core/numeric.py in isclose(a, b, rtol, atol, equal_nan)
   2519     y = array(y, dtype=dt, copy=False, subok=True)
   2520 
-> 2521     xfin = isfinite(x)
   2522     yfin = isfinite(y)
   2523     if all(xfin) and all(yfin):

TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```
works with
`ds = xr.tutorial.load_dataset('air_temperature')['air']` because no multi-dim coords?

My current work-around when coords are lost:
[...]

PR #85, opened 2019-11-15
AS_keep_all_coords

aaronspring:
- keep all coordinates, especially multi-dim coords
- fix: https://github.com/xgcm/xrft/issues/82

### DIFF
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -740,7 +740,7 @@ def test_spacing_tol_float_value(test_data_1d):
 @pytest.mark.parametrize("dim", ["time"])
 def test_keep_coords(func, dim):
     """Test whether xrft keeps multi-dim coords from rasm sample data."""
-    ds = xr.tutorial.load_dataset("rasm").fillna(0.0)["Tair"]
+    ds = xr.tutorial.load_dataset("rasm",decode_times=False).fillna(0.0)["Tair"]
     ds["time"] = np.arange(ds.time.size)
     ps = getattr(xrft, func)(ds, dim=dim)
     # check that all coords except dim from ds are kept in ps
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -735,3 +735,14 @@ def test_spacing_tol_float_value(test_data_1d):
     da = test_data_1d
     with pytest.raises(TypeError):
         xrft.dft(da, spacing_tol='string')
+
+@pytest.mark.parametrize("func", ("dft", "power_spectrum"))
+@pytest.mark.parametrize("dim", ["time"])
+def test_keep_coords(func, dim):
+    """Test whether xrft keeps multi-dim coords from rasm sample data."""
+    ds = xr.tutorial.load_dataset("rasm").fillna(0.0)["Tair"]
+    ds["time"] = np.arange(ds.time.size)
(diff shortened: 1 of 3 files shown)


xrft/xrft.py
```diff
@@ -339,6 +343,10 @@ def dft(da, spacing_tol=1e-3, dim=None, real=None, shift=True, detrend=None,
         newdims[anum] = prefix + d
 
     newcoords = {}
+    # keep former coords
+    if len(da.coords) > 1:
+        for c in da.drop(dim).coords:
+            newcoords[c] = da[c]
```
rabernat: 👍  This is the right way.

xrft/tests/test_xrft.py
```diff
@@ -735,3 +735,14 @@ def test_spacing_tol_float_value(test_data_1d):
     da = test_data_1d
     with pytest.raises(TypeError):
         xrft.dft(da, spacing_tol='string')
+
+@pytest.mark.parametrize("func", ("dft", "power_spectrum"))
+@pytest.mark.parametrize("dim", ["time"])
+def test_keep_coords(func, dim):
+    """Test whether xrft keeps multi-dim coords from rasm sample data."""
+    ds = xr.tutorial.load_dataset("rasm",decode_times=False).fillna(0.0)["Tair"]
+    ds["time"] = np.arange(ds.time.size)
```
rabernat: Why do we not decode times and overwrite the time coordinate here? Why can't we just use the dataset in original form?

rabernat:
Thinking about this a bit further, perhaps we don't want to use the xarray test data for this. It requires the network to download the data.

Instead, we could just make up a synthetic dataset or add coordinates to one of the existing ones. That would be more transparent and self contained.

xrft/tests/test_xrft.py
aaronspring: I overwrite the time because time is not evenly spaced. it is monthly datetime and fails therefore.
during travis CI testing, there was an error involving `cftime`. First I added `cftime` to `setup.py`. Then I hoped to ignore this step.

xrft/xrft.py
```diff
@@ -248,6 +248,10 @@ def dft(da, spacing_tol=1e-3, dim=None, real=None, shift=True, detrend=None,
     if not isinstance(spacing_tol, float):
         raise TypeError("Please provide a float argument")
 
+    # check for xr.da input
+    if not isinstance(da, xr.DataArray):
+        raise TypeError("Please provide xr.DataArray, found", type(da))
+
```
rabernat: Looks like this addition is not covered by tests. That's why the coverage is decreasing with PR. Could you either remove it or add a test so we can keep coverage high?

aaronspring:
Shouldn’t in theory the testing folder be excluded from the percentage score in coverage?

Implemented and tested a check whether da is xr dataarray, because I didn’t understand the error which was raise when submitting a xr dataset

rabernat:
> Shouldn’t in theory the testing folder be excluded from the percentage score in coverage?

Yes, and it is:
https://github.com/xgcm/xrft/blob/43835e11ba020c551b8adddc0cf86f68ecabf1e0/.coveragerc#L11-L14

The code in question is in `xrft.py`, not the testing folder. You added a new error check, so we needed a test for that check.

Thanks a lot for this contribution!

Issue #78, opened 2019-08-01
dft fails with real option after xarray update


tomchor:
I have recently updated xarray to the latest version and noticed that I can't perform a dft if one of the dimensions is set to have a real transform.

Example:

```
In [12]: xrft.dft(du.u, dim="x", shift=False)                                         
Out[12]: 
<xarray.DataArray 'fftn-e982410c8538317e81efde78e78ed80c' (itime: 30, freq_x: 160, y: 160, z: 144)>
dask.array<shape=(30, 160, 160, 144), dtype=complex128, chunksize=(2, 160, 160, 144)>
Coordinates:
  * itime           (itime) int64 167750 168000 168250 ... 174500 174750 175000
  * freq_x          (freq_x) float64 0.0 0.0025 0.005 ... -0.0075 -0.005 -0.0025
  * y               (y) float64 1.25 3.75 6.25 8.75 ... 391.2 393.8 396.2 398.8
  * z               (z) float64 -89.69 -89.06 -88.44 ... -1.562 -0.9375 -0.3125
    freq_x_spacing  float64 0.0025
```

Now if I use the the `real` option:

```
In [15]: xrft.dft(du.u, dim="x", shift=False, real="x")                               
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-15-f1b65d8824b4> in <module>
----> 1 xrft.dft(du.u, dim="x", shift=False, real="x")

~/miniconda3/envs/py36/lib/python3.6/site-packages/xrft/xrft.py in dft(da, spacing_tol, dim, real, shift, detrend, window, chunks_to_segments)
    350     if trans:
    351         enddims = [prefix + d for d in rawdims if d in dim]
--> 352         return daft.transpose(*enddims)
    353     else:
    354         return daft

~/miniconda3/envs/py36/lib/python3.6/site-packages/xarray/core/dataarray.py in transpose(self, transpose_coords, *dims)
   1642                 raise ValueError('arguments to transpose (%s) must be '
   1643                                  'permuted array dimensions (%s)'
-> 1644                                  % (dims, tuple(self.dims)))
   1645 
   1646         variable = self.variable.transpose(*dims)

ValueError: arguments to transpose (('freq_x',)) must be permuted array dimensions (('itime', 'y', 'z', 'freq_x'))
```

Thanks

TomNicholas:
[...]

Issue #83, opened 2019-11-15
trouble with real=True in dft


navidcy:
I am having some trouble using the real fft. I have a time series of temperature data:

```python
da = surf_temp_1selection.copy()
months = np.arange(da.time.size)+1
da.time.values = months
da = da.chunk({'time':None})
da = da.mean(dim={'xt_ocean', 'yt_ocean'})
da

<xarray.DataArray 'surface_temp' (time: 2520)>
dask.array<mean_agg-aggregate, shape=(2520,), dtype=float32, chunksize=(2520,), chunktype=numpy.ndarray>
Coordinates:
  * time     (time) int64 1 2 3 4 5 6 7 8 ... 2514 2515 2516 2517 2518 2519 2520
 ```

When I use `dft` without the `real=True` flag all seems good:
```python
dahat = xrft.dft(da, dim=['time'], window=True)
dahat

<xarray.DataArray 'rechunk-merge-43b3f58ebe71665aef5d349f10b36e8c' (freq_time: 2520)>
dask.array<rechunk-merge, shape=(2520,), dtype=complex128, chunksize=(2520,), chunktype=numpy.ndarray>
Coordinates:
  * freq_time          (freq_time) float64 -0.5 -0.4996 ... 0.4992 0.4996
    freq_time_spacing  float64 0.0003968
```

But when I give `real=True` I get:
```python
dahat = xrft.dft(da, dim=['time'], window=True, real=True)
dahat


---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-89-2e819cfb9956> in <module>
      6 
      7 print(da)
----> 8 dahat = xrft.dft(da, dim=['time'], window=True, real=True)
      9 print(" ");print(" ");print(" ");
     10 print(dahat)

/g/data3/hh5/public/apps/miniconda3/envs/analysis3-19.10/lib/python3.6/site-packages/xrft/xrft.py in dft(da, spacing_tol, dim, real, shift, detrend, window, chunks_to_segments)
    250 
    251     rawdims = da.dims
--> 252     da, trans = _transpose(da, real)
    253     if dim is None:
    254         dim = list(da.dims)

/g/data3/hh5/public/apps/miniconda3/envs/analysis3-19.10/lib/python3.6/site-packages/xrft/xrft.py in _transpose(da, real, trans)
    194         transdim = list(da.dims)
    195         if real not in transdim:
--> 196             raise ValueError("The dimension along real FT is taken must "
[...]

PR #88, opened 2019-11-19
Update 'real' argument for xarray

roxyboy:
I think this should fix issues #78 and #83 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ notifications:
 
 python:
   - 2.7
-  - 3.6
+  - 3.7
 
 env:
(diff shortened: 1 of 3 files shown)


navidcy:
@roxyboy, seems to be working fine now.

roxyboy:
Can I merge this @rabernat ?

rabernat:
Sorry for letting this hang for so long without any action.

Every time we fix a bug, we need to add a regression test. So I will merge this as soon as you add a test for the issues raised in #78 and #83.

.travis.yml
```diff
@@ -17,7 +17,7 @@ notifications:
 
 python:
   - 2.7
-  - 3.6
+  - 3.7
```
rabernat: Keep 3.6 please but also add 3.7.

navidcy:
@roxyboy could you please publish a new patch release of the package?

PR #89, opened 2020-02-08
isospectrum debugging / refactoring

apatlpo:
This PR attemps to fix several issues related to isospectrum computations:

- breaks with dask arrays (dask bincount does not handle mixed types arguments)

- breaks with extra dimensions

See [gist](https://gist.github.com/apatlpo/54f61a697d426cd5522ed1d18db66f13) for illustration of failures

I also propose a modest refactoring which consists in adding an isotropize methods that average spectra or crossspectra azimuthaly.
This method is called by isotropic_powerspectrum and isotropic_crossspectrum

There are still some improvements that could be made:

-  isotropize could be made more robust
-  the present form is not lazy (because dask bincount is not able to predict array size)

If you believe this PR is worth anything, let me know, I can finetune, even though weeks to come are going to be busy with Ocean Sciences and all.

cheers

### DIFF
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -652,14 +652,60 @@ def _synthetic_field(N, dL, amp, s):
     theta = np.fft.ifft2(np.fft.ifftshift(F_theta))
     return np.real(theta)
 
-def test_isotropize(Nx=512, Ny=256, chunkx=128, chunky=128):
-    """Test the isotropization of a power spectrum."""
-    # generate synthetic 2D spectrum, isotropize and check values
-    # check options in isotropize
-    #ps_iso = xrft.isotropize(ps, fftdim, nfactor=4)
-    # test with and without chunks
-    pass
+def _synthetic_field_xr(N, dL, amp, s,
+                    other_dim_sizes=None, dim_order=True, 
+                    chunks=None):
     
+    theta = xr.DataArray(_synthetic_field(N, dL, amp, s),
+                        dims=['y', 'x'],
+                        coords={'y':range(N), 'x':range(N)}
+                        )
+
+    if other_dim_sizes:
+        _da = xr.DataArray(np.ones(other_dim_sizes),
+                           dims=['d%d'%i for i in range(len(other_dim_sizes))])
+        if dim_order:
+            theta = theta + _da
+        else:
+            theta = _da + theta
+            
+    if chunks:
+        theta = theta.chunk(chunks)
+
+    return theta
+
+def test_isotropize(N=512):
+    """Test the isotropization of a power spectrum."""
+
+    # generate synthetic 2D spectrum, isotropize and check values
+    dL, amp, s = 1., 1e1, -3.
+    dims = ['x','y']
+    fftdim = ['freq_' + d for d in dims]    
+    def _test(da):
+        spacing_tol = 1e-3
+        nfactor = 4
+        ps = xrft.power_spectrum(da, spacing_tol, dim=dims)
+        ps_iso = xrft.isotropize(ps, fftdim, nfactor=nfactor)    
+    # np data
+    theta = _synthetic_field_xr(N, dL, amp, s)
+    _test(theta)
+    # np with other dim
+    theta = _synthetic_field_xr(N, dL, amp, s, 
+                                other_dim_sizes=[10],
+                                dim_order=True) 
+    _test(theta)
+    # da chunked, order 1
+    theta = _synthetic_field_xr(N, dL, amp, s, 
+                                chunks={'y': None, 'x': None, 'd0': 2}, 
+                                other_dim_sizes=[10], 
+                                dim_order=True) 
+    _test(theta)
+    # da chunked, order 2
+    theta = _synthetic_field_xr(N, dL, amp, s, 
+                                chunks={'y': None, 'x': None, 'd0': 2}, 
+                                other_dim_sizes=[10], 
+                                dim_order=False) 
+    _test(theta)
 
 def test_isotropic_ps_slope(N=512, dL=1., amp=1e1, s=-3.):
     """Test the spectral slope of isotropic power spectrum."""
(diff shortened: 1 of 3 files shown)


rabernat:
Hi @apatlpo -- thanks a lot for the PR! I think this makes a lot of sense, both in terms of fixing bugs
(breaks with extra dimensions is #9, a long standing issue) and code simplicity.

Can you just have a look at the test failure?

apatlpo:
Sounds good, I'll try clean up the PR.

apatlpo:
I'd like to propose renaming `isotropic_powerspectrum` and `isotropic_crossspectrum`.
Here are some suggestions:

- `isotropic_power_spectrum`, `isotropic_cross_spectrum`
- `isopower_spectrum`, `isocross_spectrum`

My preference goes to the second option.
What do people think?

apatlpo:
A test for isotropize should be added

rabernat:
> I'd like to propose renaming `isotropic_powerspectrum` and `isotropic_crossspectrum`.
> Here are some suggestions:

I'm ok with this, but let's do a deprecation cycle. We will keep the old function names for now but just have them call the new functions and issue a `DeprecationWarning`. Then in a future release we will remove them completely.

I personally prefer isotropic_power_spectrum and isotropic_cross_spectrum because they are more explicit.

apatlpo:
Sounds good. 

I won't be able to tackle before approx 1 week unfortunately.

apatlpo:
this PR has been moving along and close to be reviewed.

Before then, do I address codecov failings?

I am not sure what I'm supposed to do with these.

apatlpo:
I was about to say that this was ready for review but got an issue with python 2.7 test.
Do I really need to install 2.7 and figure what's wrong?

ping @rabernat

roxyboy:
I'm in favor of stopping support for py27...

roxyboy:
If were going to keep support py27 as for now, one way would be to check the python version on the run with `if sys.version_info[0] < 3:`.

rabernat:
Let's just drop py2.7. We don't have enough users for a deprecation cycle.

@apatlpo - just remove the py2.7 environment from the travis config and let's see if the tests go green.

xrft/tests/test_xrft.py
```diff
@@ -652,13 +652,68 @@ def _synthetic_field(N, dL, amp, s):
     theta = np.fft.ifft2(np.fft.ifftshift(F_theta))
     return np.real(theta)
 
+def _synthetic_field_xr(N, dL, amp, s,
```
rabernat: You don't need to use private functions in tests. It's not part of public API anyway. Rename to `synthetic_field`.

There should also probably be a fixture.

xrft/tests/test_xrft.py
```diff
@@ -652,13 +652,68 @@ def _synthetic_field(N, dL, amp, s):
...
+
+    # generate synthetic 2D spectrum, isotropize and check values
+    dL, amp, s = 1., 1e1, -3.
+    dims = ['x','y']
+    fftdim = ['freq_' + d for d in dims]    
+    def _test(da):
+        spacing_tol = 1e-3
+        nfactor = 4
+        ps = xrft.power_spectrum(da, spacing_tol, dim=dims)
+        ps_iso = xrft.isotropize(ps, fftdim, nfactor=nfactor)
```
rabernat: This test is not actually checking anything. It's just verifying that the function does not raise an error.

Here's an idea: rather than calling `power_spectrum`, can you just make up a synthetic spectrum with a known isotropic spectrum, and then verify that `isotropize` does the right thing? You will want some `assert` statements in there somewhere.

xrft/xrft.py
```diff
@@ -688,35 +718,19 @@ def isotropic_powerspectrum(da, spacing_tol=1e-3, dim=None, shift=True,
...
-        else:
-            newcoords[d] = k_coords[d]
-
-    return xr.DataArray(iso_ps, dims=newdims, coords=newcoords)
+def isotropic_crossspectrum(*args, **kwargs):
+    """ 
+    Deprecated function. See isotropic_cross_spectrum doc
+    """
+    import warnings
+    msg = "Deprecated. Use isotropic_cross_spectrum instead"
```
rabernat: What about "This function has been renamed and will disappear in the future. Please use isotropic_cross_spectrum instead"

xrft/tests/test_xrft.py
apatlpo: I'm not sure what a fixture precisely is, but I am under the impression that because I am calling the method multiple times with different input parameters it may not be suited to this case. Correct?

apatlpo:
Travis goes through but not codecov. 
Let me know if I need to do anything else.

rabernat:
Codecov is telling us that the test coverage has gone slightly down. The coverage tells you which percentage of the code base is covered by tests, so this means there are some new code paths that are not tested. You can use the codecov site to figure out where exactly these are. If you can add some tests 

This may seem like a pain, but it's very useful to have high test coverage like we have now, and we want to try to maintain this. You can check coverage locally while developing by running this command:
https://github.com/xgcm/xrft/blob/41cf8069033c989cf580eab5c5fb67dc3ce61e20/.travis.yml#L47

This should be a relatively simple fix, but it turns out to be complicated or annoying, we can just  accept the drop in coverage and move on.

Thanks for your perseverance with this PR.

apatlpo:
thanks for the explanation Ryan.

I had to add tests for the methods that will be soon deprecated.

apatlpo:
Ping @rabernat @roxyboy
All lights are green

rabernat:
> I had to add tests for the methods that will be soon deprecated.

Ah ok, I understand. In that case, please add a `# pragma: no cover` comment (https://coverage.readthedocs.io/en/coverage-4.3.3/excluding.html) to the deprecated methods and then remove the redundant tests.

Then we are done.

apatlpo:
done

rabernat:
Fantastic! Thanks so much for your persistence! We really appreciate having you as a contributor to xrft.

PR #90, opened 2020-02-15
Refactoring new dims and coords, and deprecation warning for Python 2

roxyboy:


rabernat:
I was thinking we could do something like this:

```python
def freq_for_dim(Nx, dx, real, shift):
    # figure out frequency
    return k
```


```

**Comment by roxyboy** [comment]
@lanougue Do you think you can incorporate `shift_phase` in the `_freq` function?

File: xrft/xrft.py [hunk 1]
```diff
@@ -202,9 +203,52 @@ def _transpose(da, real, trans=False):
...
+        # consistent with np.fft.rfftn
+        fftfreq = [np.fft.fftfreq]*(len(N)-1)
+        fftfreq.append(np.fft.rfftfreq)
+
+    k = [fftfreq(Nx, dx) for (fftfreq, Nx, dx) in zip(fftfreq, N, delta_x)]
+
+    if shift:
+        k = [np.fft.fftshift(l) for l in k]
+
+    return k
```
rabernat: Ok so now we have the frequency logic all by itself. @lanougue - do you think there is something here that needs to be fixed?

lanougue: It seems ok. We can leave the shifting of thecoords  outside of this function (for ifft)

lanougue: When defining newdims names in  _new_dims_and_coords() function, do you want to keep something as 'freq_freq_x' as a possibility ?
I my PR, I proposed something as
'''
newdims[anum] = prefix + d if d[:len(prefix)]!=prefix else d[len(prefix):]
'''
which will transform back 'freq_x' into 'x'

roxyboy: I agree that we should probably avoid the possibility such as `freq_freq_x`. The `prefix` argument in `dft` should make this flexible?

roxyboy: I adopted your code: `newdims[anum] = prefix + d if d[:len(prefix)]!=prefix else d[len(prefix):]`

roxyboy: @lanougue Are you suggesting that we incorporate the shifting of the coords for ifft for the next round of PR after merging this one? I think we should incorporate your enhancements!

lanougue: No, what I mean is that shifting the coords in ifft should be done outside the _freq() function.
It definitively pertains to ifft() function

rabernat:
Ok, LGTM!

lanougue:
Reopening this point.

If original coordinate name is something like "freq_x", applying xrft.dft return a "x" coordinate, which is the expected behaviour.
However, the spacing attribute is not correctly named. Instead of "x_spacing" we get "freq_freq_x_spacing.

Since @roxyboy was in charge of this modification, I let him correct this.
cheers

lanougue:
I did a new PR Phase corrections #122  to solve other things with the forward transform.
I finally also added corrections mentioned above about naming new coordinates and spacing in _new_dims_and_coords() function.
@rabernat , @roxyboy , if you can take a look

PR #91, opened 2020-02-16
Simpler examples

roxyboy:


roxyboy:
@rabernat Let me know if the mooring example is good?

rabernat:
Thanks for getting this started. However, I was hoping we could use current meter data, which would should the inertial frequency, tidal frequencies, etc.

rabernat:
Thanks for getting this started. However, I was hoping we could use current meter data, which would should the inertial frequency, tidal frequencies, etc.

rabernat:
Thanks for getting this started. However, I was hoping we could use current meter data, which would should the inertial frequency, tidal frequencies, etc.

rabernat:
Thanks for getting this started. However, I was hoping we could use current meter data, which would should the inertial frequency, tidal frequencies, etc.

navidcy:
Perhaps the first example could be a synthetic single? E.g., construct a signal as a sum of two frequencies and use `xrft.dft` to compute the spectrum and compare with what you actually expect?

rabernat:
Great suggestion @navidcy!

navidcy:
@roxyboy, is there a way to see a preview of the docs after, e.g., this PR?

PR #93, opened 2020-04-13
Fix dependencies listed under install_requires

shoyer:
I'm excited to give xrft a try!

I noticed that future and docrep are no longer imported in xrft.py, but pandas
and scipy are, so this PR updates dependencies accordingly in setup.py.


rabernat:
Thanks for this @shoyer! Our packaging is definitely a bit sloppy.

Let us know if you have any questions or feedback. There's lots of room for improvement.

PR #96, opened 2020-04-17
raise a warning when coordinates are in lat/lon

roxyboy:
PR to raise a warning when the coordinates to take the FFT over are in latitude or longitude coordinates so as to notify the user that the inferred wavenumbers won't have MKS units.

### DIFF
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -2,6 +2,7 @@ import numpy as np
 import pandas as pd
 import xarray as xr
 import dask.array as dsar
+import warnings
 
 import scipy.signal as sps
 import scipy.linalg as spl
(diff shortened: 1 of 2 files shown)


rabernat:
If you'd like someone to review this PR, I recommend writing a description.

dcherian:
I feel like a warning is not necessary. A simple note in the docstring would work. 

(This warning wouldn't be raised if the dimension names were lat/long so it isn't the most effective tactic.)

roxyboy:
@dcherian Thanks for you feedback. While I understand your point, putting it in the docstring would require to have the same comment in each function of xrft. Since all of the functions depend on the `dft` function, the idea was that raising a warning in `dft` would work universally within xrft no matter which function the user calls. 

The focus on lat/long was that they are very common in the earth sciences but I am open to changing this.

Issue #99, opened 2020-07-28
CFTime vs. pd.datetime


ks905383:
Found out the type error issue I mentioned in the other thread. 

The dataset I was using has time imported `cftime` format (since it's a `noleap` calendar, which are imported as `cftime.datetime` index per `xarray` docs). 

Unfortunately, `cftime` formatted dates cause an error when `dft` is run on the `time` dimension, in the `# verify even spacing of input coordinates` section:

```
for d in dim:
        coord = da[d]
        diff = np.diff(coord)
        if pd.api.types.is_timedelta64_dtype(diff):
            # convert to seconds so we get hertz
            diff = diff.astype('timedelta64[s]').astype('f8')
        delta = diff[0]
        if not np.allclose(diff, diff[0], rtol=spacing_tol):
            raise ValueError("Can't take Fourier transform because "
                             "coodinate %s is not evenly spaced" % d)
        delta_x.append(delta)
```

Specifically, the `if` statement only converts `timedelta64` type values to integers; in my case, `diff` is kept as `datetime.timedelta`. Since `np.allclose()` requires numeric inputs, the code fails at `if not np.allclose(diff,diff[0],rtol=spacing_tol)`. 

**Dealing more generally with `noleap` calendars**
A related issue is that even converting `diff` to a supported format (using `ds.indexes['time'].to_datetimeindex()` poses a problem - 

```
/home/kschwarzwald/.conda/envs/climate1/lib/python3.7/site-packages/ipykernel_launcher.py:1: RuntimeWarning: Converting a CFTimeIndex with dates from a non-standard calendar, 'noleap', to a pandas.DatetimeIndex, which uses dates from the standard calendar.  This may lead to subtle errors in operations that depend on the length of time between dates.
  """Entry point for launching an IPython kernel.
```
In other words, `diff` will now have twice the value on "leap" days, which don't exist in the original dataset, and therefore trigger the "coordinate is not evenly spaced error" (and presumably cause issues with the Fourier transform).

**Proposed Solution** 
A first useful addition would probably be raising a `TypeError` on `cftime` dimensions. 

But further, I think this can be fixed by just adding something like:

```
if type(coord.values[0]) is cftime._cftime.DatetimeNoLeap:
  coord = cftime.date2num(coord,'hours since 1800-01-01 00:00:00','noleap')
```
before `diff` is called, which puts it in a numeric format that's readable by `np.allclose`. 

**Example of Error** 
```
ds = xr.open_dataset('test.nc')
[...]

Issue #97, opened 2020-07-28
Undocumented error on dimension names


ks905383:
I've been having issues running `xrft` (0.2.0) - an error of `'t' not found in array dimensions ('time', 'lat', 'lon')` appears, and I haven't been able to diagnose what's causing it. 

Here's an example. Running the following causes the error (using the attached file, which is just a `time x lat x lon` precipitation dataset):
```
ds = xr.open_dataset('test.nc')
xrft.power_spectrum(ds.pr.chunk({'lat':1,'lon':1}),dim='time',detrend='linear',window=True)
```

It doesn't seem to be related to the input options though - the same error shows up just when running `xrft.power_spectrum(ds.pr,dim='time')`. `t` is not a dimension in the dataset (and at least from the docs, it doesn't look like a dimension has to be named that either?) 

Apologize if this is not a bug and I'm just missing something (and that I haven't been able to get more info on what's causing this), but in that case it might be nice to have a specific error message for it. 

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
~/.conda/envs/climate1/lib/python3.7/site-packages/xarray/core/common.py in _get_axis_num(self, dim)
    172         try:
--> 173             return self.dims.index(dim)
    174         except ValueError:

ValueError: tuple.index(x): x not in tuple

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
<ipython-input-46-44b7f9b05769> in <module>
----> 1 xrft.power_spectrum(ds.pr.chunk({'lat':1,'lon':1}),dim='time',detrend='linear',window=True)

~/.conda/envs/climate1/lib/python3.7/site-packages/xrft/xrft.py in power_spectrum(da, spacing_tol, dim, real, shift, detrend, window, chunks_to_segments, density)
    400     daft = dft(da, spacing_tol,
    401               dim=dim, real=real, shift=shift, detrend=detrend, window=window,
--> 402               chunks_to_segments=chunks_to_segments)
    403 
    404     if dim is None:

~/.conda/envs/climate1/lib/python3.7/site-packages/xrft/xrft.py in dft(da, spacing_tol, dim, real, shift, detrend, window, chunks_to_segments)
    280 
    281     # the axes along which to take ffts
--> 282     axis_num = [da.get_axis_num(d) for d in dim]
    283 
    284     N = [da.shape[n] for n in axis_num]

[...]

PR #100, opened 2020-07-29
cftime convention for noleap

roxyboy:
I added some checks to the `cftime` convention based on #99 but I don't have much confidence whether this is robust enough for other options than `noleap`. If others think this PR is good to merge, I'll merge it and work on releasing a new version #97.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ CLASSIFIERS = [
     'Topic :: Scientific/Engineering',
 ]
 
-INSTALL_REQUIRES = ['xarray', 'dask', 'numpy', 'pandas', 'scipy']
+INSTALL_REQUIRES = ['xarray', 'dask', 'numpy', 'pandas', 'scipy','cftime']
 SETUP_REQUIRES = ['pytest-runner']
 TESTS_REQUIRE = ['pytest >= 2.8', 'coverage']
 
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,8 @@ CLASSIFIERS = [
     'Topic :: Scientific/Engineering',
 ]
 
-INSTALL_REQUIRES = ['xarray', 'dask', 'numpy', 'pandas', 'scipy','cftime']
+INSTALL_REQUIRES = ['xarray', 'dask', 'numpy', 'pandas', 'scipy']
+EXTRAS_REQUIRE = ['cftime']
 SETUP_REQUIRES = ['pytest-runner']
 TESTS_REQUIRE = ['pytest >= 2.8', 'coverage']
 
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -198,6 +198,15 @@ class TestDFTImag(object):
         freq_time_expected = np.fft.fftfreq(Nt, dt)
         npt.assert_allclose(ft['freq_time'], freq_time_expected)
 
+        time = cftime.num2date(np.arange(0,10*365), units, '360_day')
+        Nt = len(time)
+        da = xr.DataArray(np.arange(len(time)), dims='time',
+                                    coords=[time])
+        ft = xrft.dft(da, shift=False)
+        dt = np.diff(time)[0].total_seconds()
+        freq_time_expected = np.fft.fftfreq(Nt, dt)
+        npt.assert_allclose(ft['freq_time'], freq_time_expected)
+
     def test_dft_2d(self):
         """Test the discrete Fourier transform on 2D data"""
         N = 16
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -1,6 +1,7 @@
 import numpy as np
 import pandas as pd
 import xarray as xr
+import cftime
 import dask.array as dsar
 
 import scipy.signal as sps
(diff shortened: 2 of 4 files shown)


roxyboy:
@rabernat Is it ok for me to merge this??

rabernat:
> @rabernat Is it ok for me to merge this??

No. Someone needs to review your PR. I appreciate all your work on this. Please try to be patient.

xrft/xrft.py
```diff
@@ -249,6 +250,21 @@ def _new_dims_and_coords(da, axis_num, dim, wavenm, prefix):
 
     return newdims, newcoords
 
+def _cftime(coord):
```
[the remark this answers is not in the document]
rabernat: And let's define clearly whether this function should return a DataArray or numpy array. Currently it might return either.

setup.py
```diff
@@ -22,7 +22,7 @@
     'Topic :: Scientific/Engineering',
 ]
 
-INSTALL_REQUIRES = ['xarray', 'dask', 'numpy', 'pandas', 'scipy']
+INSTALL_REQUIRES = ['xarray', 'dask', 'numpy', 'pandas', 'scipy','cftime']
```
rabernat, line 25: How about we make cftime an optional dependency?

If a user ever passes a cftime datatype to xrft, we know they already have cftime installed.

xrft/tests/test_xrft.py
```diff
@@ -1,6 +1,7 @@
 import numpy as np
 import pandas as pd
 import xarray as xr
+import cftime
```
rabernat, line 4: We can move this import inside the `_maybe_decode_time` function to avoid the hard dependency.

xrft/xrft.py
```diff
@@ -341,7 +357,7 @@ def dft(da, spacing_tol=1e-3, dim=None, real=None, shift=True, detrend=None,
     # verify even spacing of input coordinates
     delta_x = []
     for d in dim:
-        coord = da[d]
+        coord = _cftime(da[d])
```
rabernat, line 360: As is `_cftime` returns a numpy array, but `coord` is a DataArray. I don't like the idea of changing the type of `coord` like this. I recommend putting all the decoding logic inside `_maybe_decode_time`.

xrft/xrft.py
```diff
@@ -249,6 +250,21 @@ def _new_dims_and_coords(da, axis_num, dim, wavenm, prefix):
...
+    elif type(coord.values[0]) == cftime._cftime.DatetimeGregorian:
+        coord = cftime.date2num(coord,'seconds since 1800-01-01 00:00:00',
+                                'standard')
+    elif type(coord.values[0]) == cftime._cftime.DatetimeJulian:
+        coord = cftime.date2num(coord,'seconds since 1800-01-01 00:00:00',
+                                'julian')
+    elif type(coord.values[0]) == cftime._cftime.Datetime360Day:
+        coord = cftime.date2num(coord,'seconds since 1800-01-01 00:00:00',
+                                '360_day')
+    return coord
```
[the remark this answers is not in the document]
roxyboy: I wasn't able to find a way to moving everything under `pd.api.types.is_timedelta64_dtype()` as for `cftime`, we need to decode the time before taking the difference. I could change the logic to use `pd.api.types.is_datetime64_dtype()` but it ended up messy either way.

xrft/tests/test_xrft.py
```diff
@@ -39,6 +40,14 @@ def test_data_1d(request):
         da = da.chunk()
     return da
 
+@pytest.fixture(params=['pandas','standard','julian','365_day','360_day'])
+def test_time(request):
```
rabernat, line 44: Fixtures should not have the word `test` in the function name.

xrft/xrft.py
```diff
@@ -249,6 +249,17 @@ def _new_dims_and_coords(da, axis_num, dim, wavenm, prefix):
 
     return newdims, newcoords
 
+def _maybe_decode_time(coord):
+    """Returns decoded time as a xarray.DataArray."""
+    import cftime
```
rabernat, line 254: As is, this function will fail if cftime is not installed. So we need to be a little sneakier. This import should only happen if `coord` is a cftime type. Maybe something like
```python
v0 = coord.values[0]
calendar = getattr(v0, 'calendar', None) 
if v0:
    import cftime
    decoded_time = cftime.date2num(coord, ref_units, calendar)
```
```

File: xrft/xrft.py (hunk 5 above)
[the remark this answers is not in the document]
Comment (rabernat) [reply xrft/xrft.py:261]: As is, this function will fail if cftime is not installed. So we need to be a little sneakier. This import should only happen if `coord` is a cftime type. Maybe something like
```python
v0 = coord.values[0]
calendar = getattr(v0, 'calendar', None) 
if calendar:
    import cftime
    decoded_time = cftime.date2num(coord, ref_units, calendar)
elif api.types.is_datetime64_any_dtype(v0):
    decoded_time = # something to decode regular datetime
```

That would solve your pandas problem too.

xrft/tests/test_xrft.py
rabernat, line 44: Fixtures should not have the word `test` in the function name.

xrft/xrft.py
```diff
@@ -342,6 +353,10 @@ def dft(da, spacing_tol=1e-3, dim=None, real=None, shift=True, detrend=None,
     delta_x = []
     for d in dim:
         coord = da[d]
+        try: # check for cftime coordinates
+            coord = _maybe_decode_time(coord)
+        except:
```
rabernat, line 358: It's bad form to except any exception type. What are you trying to catch? I don't see why a try block is needed here at all.  Why not have `_maybe_decode_time` pass the coordinate through unchanged if it's not a time type.

xrft/xrft.py
```diff
@@ -249,6 +249,23 @@ def _new_dims_and_coords(da, axis_num, dim, wavenm, prefix):
 
     return newdims, newcoords
 
+def _maybe_decode_time(coord):
```
rabernat, line 252:
```suggestion
def _diff_coord(coord):
```

xrft/xrft.py
```diff
@@ -341,11 +358,7 @@ def dft(da, spacing_tol=1e-3, dim=None, real=None, shift=True, detrend=None,
     # verify even spacing of input coordinates
     delta_x = []
     for d in dim:
-        coord = da[d]
-        diff = np.diff(coord)
-        if pd.api.types.is_timedelta64_dtype(diff):
-            # convert to seconds so we get hertz
-            diff = diff.astype('timedelta64[s]').astype('f8')
+        diff = _maybe_decode_time(da[d])
```
rabernat, line 361:
```suggestion
        diff = _diff_coord(da[d])
```

PR #101, opened 2020-07-31
Attempt to automate release

roxyboy:
I started working on the process to automate package distribution releases (following [this](https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/)) but... I'm new to this so any feedback would be much appreciated.


rabernat:
I recommend copying what we do in xgcm:

https://github.com/xgcm/xgcm/blob/master/.github/workflows/pythonpublish.yaml

roxyboy:
@rabernat I copied the xgcm yaml file. Does this now mean that if I merge this and release a new version, it will automatically be pushed to pypi?

.github/workflows/pythonpublish.yaml
```diff
@@ -0,0 +1,27 @@
...
+        with:
+          python-version: '3.x'
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install setuptools setuptools-scm wheel twine
+      - name: Build and publish
+        env:
+          TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
+          TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
```
rabernat, line 23: Have you set the secrets on this repo?

rabernat: We should be able to get the secrets at the organization level.

rabernat:
Let's merge this and see if it works!

PR #103, opened 2020-08-26
Fix normalisation of power spectra density when `chunks_to_segments=True`

### LINKED ISSUE #102: Bug when using `density=True` with `chunks_to_segments=True`
I think there is a bug when using `density=True` with `chunks_to_segments=True`. This can be demonstrated by testing Parseval's theorem for the case of chunked arrays. 

For an unchunked array, the integral of the psd matches the variance:
```python
import numpy as np
import xarray as xr
from xrft.xrft import power_spectrum

data = xr.DataArray(np.random.normal(0, 1, size=1000), 
                    dims=['t'], 
                    coords=[range(0,1000)])

psd = power_spectrum(data, dim=['t'])

print(f"Integral of psd: {psd.integrate('freq_t').values}")
print(f"Variance: {(data.std()**2).values}")
```
```
Integral of psd: 0.9578122989216458
Variance: 0.9577984563349068
```

However, if we chunk the array into more than one chunk, the integrals of the psd for each segment do not match the corresponding variances:
```python
data_chunked = data.chunk({'t': 500})

psd = power_spectrum(data_chunked, dim=['t'], 
                     chunks_to_segments=True)

print(f"Integral of psd: {psd.integrate('freq_t').values}")
print(f"Variance: {[(data[:500].std()**2).item(), (data[500:].std()**2).item()]}")
```
```
Integral of psd: [0.2404514 0.2382441]
Variance: [0.958114994451872, 0.9466505378063875]
```

This is because the `N` used in the normalisation with `density=True` is computed from the dimension sizes of the input array rather than from the chunk sizes (so in the example above, the powers are off by a factor of (1000/500)**2 = 4).

dougiesquire:
# Description

The term `N` in the `density=True` normalisation is computed from the chunk sizes rather than dimension sizes when `chunks_to_segments=True`.

Closes #102

## Type of change

-   [x]  Bug fix (non-breaking change which fixes an issue)

## Testing 

- [x] Extended `test_parseval()` to include checks with `chunks_to_segments=True`

### DIFF
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -548,51 +548,95 @@ class TestCrossPhase(object):
         npt.assert_almost_equal(actual_phase_offset, phase_offset)
 
 
-def test_parseval():
+@pytest.mark.parametrize("chunks_to_segments", [False, True])
+def test_parseval(chunks_to_segments):
     """Test whether the Parseval's relation is satisfied."""
 
-    N = 16
+    N = 16 # Must be divisible by n_segments (below)
     da = xr.DataArray(np.random.rand(N,N),
                     dims=['x','y'], coords={'x':range(N), 'y':range(N)})
     da2 = xr.DataArray(np.random.rand(N,N),
                     dims=['x','y'], coords={'x':range(N), 'y':range(N)})
+    
+    if chunks_to_segments:
+        n_segments = 2
+        # Chunk da and da2 into n_segments
+        da = da.chunk({'x': N / n_segments, 'y': N / n_segments})
+        da2 = da2.chunk({'x': N / n_segments, 'y': N / n_segments})
+    else:
+        n_segments = 1
 
     dim = da.dims
+    fftdim = [f'freq_{d}' for d in da.dims]
     delta_x = []
     for d in dim:
         coord = da[d]
         diff = np.diff(coord)
         delta = diff[0]
         delta_x.append(delta)
+    delta_xy = np.asarray(delta_x).prod() # Area of the spacings
+    
+    ### Test Parseval's theorem for power_spectrum with `window=False` and detrend=None
+    ps = xrft.power_spectrum(da, 
+                             chunks_to_segments=chunks_to_segments)
+    # If n_segments > 1, use xrft._stack_chunks() to stack each segment along a new dimension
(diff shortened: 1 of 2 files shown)


dougiesquire:
Ready for review @rabernat

xrft/xrft.py
```diff
@@ -453,7 +453,11 @@ def power_spectrum(da, spacing_tol=1e-3, dim=None, real=None, shift=True,
     # the axes along which to take ffts
     axis_num = [da.get_axis_num(d) for d in dim]
 
-    N = [da.shape[n] for n in axis_num]
+    if chunks_to_segments:
+        # Use chunk sizes for normalisation
+        N = [da.chunks[n][0] for n in axis_num]
```
rabernat, line 458: Since this code block  is re-used in two places functions, maybe we could put it in a small standalone function, with the signature

```python
def _calc_normalization_factor(da, axis_num, chunks_to_segments):
```

xrft/tests/test_xrft.py
```diff
@@ -548,51 +548,77 @@ def test_cross_phase_2d(self, dask):
...
-    ps = xrft.power_spectrum(da, window=True, detrend='constant')
-    da_prime = da.values - da.mean(dim=dim).values
-    npt.assert_almost_equal(ps.values.sum(),
+    window = xr.DataArray(
+        np.tile(
+            np.hanning(N / n_segments) * np.hanning(N / n_segments)[:, np.newaxis],
+            (n_segments, n_segments)
+        ),
+        dims=dim, coords=da.coords
+    )
```
rabernat, line 583: I guess this complex window is necessary to provide a reference "truth" against which to compare the results of xrft?

In general, I'd like to see a few comments in this test to help the other devs understand what's happening better.

rabernat (changes requested):
This is looking great! Thanks @dougiesquire! Just a few small comments / questions.

Sorry for the slow review.

xrft/tests/test_xrft.py
dougiesquire: That's right - this is to account for the energy reduction resulting from windowing (see also https://github.com/xgcm/xrft/issues/104). 

I've added some comments and made some small changes to `test_parseval` that make it clearer (at least to me...)

rabernat:
Fantastic! Thank you!

PR #107, opened 2020-09-12
Add pypi and anaconda badges

navidcy:


rabernat:
Thanks for this @navidcy!

PR #110, opened 2020-09-30
Updates Repo's Badges

navidcy:
This PR updates most of the badges that appear in the repo's README.

- Adds code black
- Uses latest version of zenodo url. According to Zenodo: "the doi:[10.5281/zenodo.1402635](https://doi.org/10.5281/zenodo.1402635) represents all versions, and will always resolve to the latest one."
- Adds the how-many-times-downloaded badge from conda-forge


navidcy:
I'll elaborate a bit on this: the current zenodo doc points to [v0.1-beta](https://doi.org/10.5281/zenodo.1402635). So either the zenodo url must be manually updated after each release, or the generic zenodo url that points to the latest release should be used. This PR suggest the latter. :)

PR #112, opened 2020-10-09
Reorganize detrending

### LINKED ISSUE #111: Issues with detrend='linear' in dft() killing kernel
I have two comments, the first of which is the main issue, while the second is a small, related note. I am currently running the most up-to-date xrft version from GitHub (version 0.2.2+1.g8b7e84e) on Pangeo Cloud.

(1)  I've been trying to use the 'linear' detrend option in `dft()` without success. (I created a sample notebook [here](https://github.com/paigem/pangeo_tests/blob/master/Test_dft_linear_detrend.ipynb) on a subset of CESM data to illustrate this problem, which should be executable on Pangeo Cloud.) Whenever I use the 'linear' option, when using `.load()` to calculate the dft function, the cell runs for awhile (~15 minutes in my notebook example) and eventually kills the kernel without any warning or error messages. This does not seem to be an issue with running the computation with dask gateway, as I have had this problem running both with and without dask (and when trying to use dask gateway, the dashboard showed no activity the whole time). When running with `detrend='constant'` or `detrend=None` I don't have any errors and the function runs in under 10 seconds. I've looked at the xrft code a bit and it may have to do with the dask chunking, but I'm not sure at this point. I believe @navidcy has had similar problems.

~(2) I just noticed today that I've been running the dft() function with `detrend=True`. This ran just fine, but also didn't raise any errors saying that 'True' is, in fact, not one of the options. I wonder if it makes sense to add a warning that any arguments that are not either 'constant' or 'linear' will be treated as 'None'? (I realize that this was entirely my mistake in not reading the xrft docs better, but since the 'window' flag does use 'True' it may be confusing for others as well.)~ (This comment is now in issue #114.)

I would be happy to make a pull request for these issues if we see the need, but since I'm new to this community I might need a few tips from the more experienced developers out there!

### LINKED ISSUE #113: Redundant check for chunked dimensions?
hey, this part here
https://github.com/xgcm/xrft/blob/8b7e84eb414ed02fff059998d2396430417a1d29/xrft/xrft.py#L128-L131
seems to be checking that the axis along FFT is to be done are not chunked.

We are wondering (with @paigem) whether this test is necessary at this point. It seems that if the user asks for FFT along chunked dimensions, `dask/array/fft.py` will raise an informative error. So perhaps we should get rid of this check above?

### LINKED ISSUE #114: Raise warning for non-supported `detrend` arguments
I noticed that I've been running the dft() function with `detrend=True`. This ran just fine, but also didn't raise any errors saying that 'True' is, in fact, not one of the options. I wonder if it makes sense to add a warning that any arguments that are not either 'constant' or 'linear' will be treated as 'None'? (I realize that this was entirely my mistake in not reading the xrft docs better, but since the 'window' flag does use 'True' it may be confusing for others as well.)

This was originally brought up in issue #111, but adding as separate issue here since it is distinct from the topic of #111.

paigem:
- Replace scipy.signal.detrend with xarray.polyfit for 1-d fft detrend. If user asks for fft in only one dimension with linear detrend, xr.polyfit detrending is called.
- Reorganize `_apply_detrend()`. 
- Raise error if non-detrended dimensions do not have chunks of length 1

Worked together with @navidcy  

Closes #111, closes #113, closes #114

### DIFF
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -499,12 +499,18 @@ class TestSpectrum(object):
         npt.assert_almost_equal(np.ma.masked_invalid(ps).mask.sum(), 0.)
 
         ### Remove least-square fit
-        ps = xrft.power_spectrum(da, dim=['y','x'],
+        if not dask:
+            with pytest.raises(ValueError):
+                ps = xrft.power_spectrum(da, dim=['y','x'],
+                                        window=True, density=False, detrend='linear'
+                                        )
+        else:
+            ps = xrft.power_spectrum(da, dim=['y','x'],
                                 window=True, density=False, detrend='linear'
                                 )
-        daft = xrft.dft(da, dim=['y','x'], window=True, detrend='linear')
-        npt.assert_almost_equal(ps.values, np.real(daft*np.conj(daft)))
-        npt.assert_almost_equal(np.ma.masked_invalid(ps).mask.sum(), 0.)
+            daft = xrft.dft(da, dim=['y','x'], window=True, detrend='linear')
+            npt.assert_almost_equal(ps.values, np.real(daft*np.conj(daft)))
+            npt.assert_almost_equal(np.ma.masked_invalid(ps).mask.sum(), 0.)
 
     @pytest.mark.parametrize("dask", [False, True])
     def test_cross_spectrum(self, dask):
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -79,18 +79,18 @@ def numpy_detrend(da):
 
 def test_detrend():
     N = 16
-    
+
     nt, nz, ny, nx = N-6, int(N/2), N-1, N+1
-    
+
     t = np.linspace(-int(N/2), int(N/2), nt)
     z = np.arange(nz)
     y = np.arange(ny)
     x = np.arange(nx)
-    
+
     # create some noise; make sure it has no trend
     def detrended_noise(N, amplitude=1.0):
         return sps.detrend(amplitude*np.random.rand(N))
-    
+
     noise_t = detrended_noise(nt, amplitude=0.12)
     noise_z = detrended_noise(nz, amplitude=0.3)
     noise_y = detrended_noise(ny, amplitude=0.2)
@@ -101,24 +101,24 @@ def test_detrend():
     data_z =  0.2*z + noise_z
     data_y = -0.3*y + noise_y
     data_x =  0.5*x + noise_x
-    
+
     # construct some 3D/4D numpy arrays to be used as tests for the detrendn function
-    array3D = (data_t[:,np.newaxis,np.newaxis] 
+    array3D = (data_t[:,np.newaxis,np.newaxis]
             + data_y[np.newaxis,:,np.newaxis]
             + data_x[np.newaxis,np.newaxis,:]
           )
 
-    array3D_detrended = (noise_t[:,np.newaxis,np.newaxis] 
+    array3D_detrended = (noise_t[:,np.newaxis,np.newaxis]
             + noise_y[np.newaxis,:,np.newaxis]
             + noise_x[np.newaxis,np.newaxis,:]
           )
-          
+
     array4D = (data_t[:,np.newaxis,np.newaxis,np.newaxis]
             + data_z[np.newaxis,:,np.newaxis,np.newaxis]
             + data_y[np.newaxis,np.newaxis,:,np.newaxis]
             + data_x[np.newaxis,np.newaxis,np.newaxis,:]
           )
-          
+
     array4D_detrendedy = (data_t[:,np.newaxis,np.newaxis,np.newaxis]
             + data_z[np.newaxis,:,np.newaxis,np.newaxis]
             + noise_y[np.newaxis,np.newaxis,:,np.newaxis]
@@ -147,39 +147,40 @@ def test_detrend():
     da4D = xr.DataArray(array4D, dims=['time','z','y','x'],
                      coords={'time':t,'z':z,'y':y,'x':x}
                      )
-    
+
     da3D_detrended = xr.DataArray(array3D_detrended,dims=['time','y','x'],
                      coords={'time':t,'y':y,'x':x}
                      )
-    
+
     da4D_detrendedy = xr.DataArray(array4D_detrendedy, dims=['time','z','y','x'],
                      coords={'time':t,'z':z,'y':y,'x':x}
                      )
-    
+
     da4D_detrendedxy = xr.DataArray(array4D_detrendedxy, dims=['time','z','y','x'],
                      coords={'time':t,'z':z,'y':y,'x':x}
                      )
-    
+
     da4D_detrendedxyz = xr.DataArray(array4D_detrendedxyz, dims=['time','z','y','x'],
                      coords={'time':t,'z':z,'y':y,'x':x}
                      )
 
     func = xrft.detrend_wrap(xrft.detrendn)
-    
+
     # let's begin testing
-    
+
     #########
     # Chunk along the `time` axis
     #########
     da = da4D.chunk({'time': 1})
     with pytest.raises(ValueError):
         func(da.data, axes=[0,1,2,3]).compute()
-    
+
     # test detrending along 1 dimension
-    da_prime = func(da, axes=[2]).compute()
+    #da_prime = func(da, axes=[2]).compute()
+    da_prime = xrft.apply_detrend(da, 'y', 2, 'linear')
     npt.assert_allclose(da_prime, da4D_detrendedy)
     npt.assert_allclose(da_prime[0,0,:,0].data, noise_y)
-    
+
     # test detrending along >1 dimensions
     da_prime = func(da.data, axes=[1,2,3]).compute()
     npt.assert_allclose(da_prime.data, array4D_detrendedxyz)
@@ -187,7 +188,7 @@ def test_detrend():
 
     da = da3D.chunk({'time': None})
     # test detrending along all dimensions (total dimensions <4)
-    da_prime = func(da.data).compute()
+    da_prime = func(da.data,axes=[0,1,2]).compute()
     npt.assert_allclose(da_prime, da3D_detrended)
 
     #########
@@ -637,7 +638,7 @@ def test_parseval(chunks_to_segments):
                     dims=['x','y'], coords={'x':range(N), 'y':range(N)})
     da2 = xr.DataArray(np.random.rand(N,N),
                     dims=['x','y'], coords={'x':range(N), 'y':range(N)})
-    
+
     if chunks_to_segments:
         n_segments = 2
         # Chunk da and da2 into n_segments
@@ -655,24 +656,24 @@ def test_parseval(chunks_to_segments):
         delta = diff[0]
         delta_x.append(delta)
     delta_xy = np.asarray(delta_x).prod() # Area of the spacings
-    
+
     ### Test Parseval's theorem for power_spectrum with `window=False` and detrend=None
-    ps = xrft.power_spectrum(da, 
+    ps = xrft.power_spectrum(da,
                              chunks_to_segments=chunks_to_segments)
     # If n_segments > 1, use xrft._stack_chunks() to stack each segment along a new dimension
     da_seg = xrft.xrft._stack_chunks(da, dim).squeeze() if chunks_to_segments else da
     da_prime = da_seg
     # Check that the (rectangular) integral of the spectrum matches the energy
-    npt.assert_almost_equal((1 / delta_xy) * ps.mean(fftdim).values, 
-                            (da_prime**2).mean(dim).values, 
+    npt.assert_almost_equal((1 / delta_xy) * ps.mean(fftdim).values,
+                            (da_prime**2).mean(dim).values,
                             decimal=5)
-    
+
     ### Test Parseval's theorem for power_spectrum with `window=True` and detrend='constant'
-    # Note that applying a window weighting reduces the energy in a signal and we have to account 
-    # for this reduction when testing Parseval's theorem. 
-    ps = xrft.power_spectrum(da, 
-                             window=True, 
-                             detrend='constant', 
+    # Note that applying a window weighting reduces the energy in a signal and we have to account
+    # for this reduction when testing Parseval's theorem.
+    ps = xrft.power_spectrum(da,
+                             window=True,
+                             detrend='constant',
                              chunks_to_segments=chunks_to_segments)
     # If n_segments > 1, use xrft._stack_chunks() to stack each segment along a new dimension
     da_seg = xrft.xrft._stack_chunks(da, dim).squeeze() if chunks_to_segments else da
@@ -686,21 +687,21 @@ def test_parseval(chunks_to_segments):
         dims=dim, coords=da.coords
     )
     # Check that the (rectangular) integral of the spectrum matches the windowed variance
-    npt.assert_almost_equal((1 / delta_xy) * ps.mean(fftdim).values, 
-                            ((da_prime*window)**2).mean(dim).values, 
+    npt.assert_almost_equal((1 / delta_xy) * ps.mean(fftdim).values,
+                            ((da_prime*window)**2).mean(dim).values,
                             decimal=5)
-    
+
     ### Test Parseval's theorem for cross_spectrum with `window=True` and detrend='constant'
-    cs = xrft.cross_spectrum(da, da2, 
-                             window=True, 
+    cs = xrft.cross_spectrum(da, da2,
+                             window=True,
                              detrend='constant',
                              chunks_to_segments=chunks_to_segments)
     # If n_segments > 1, use xrft._stack_chunks() to stack each segment along a new dimension
     da2_seg = xrft.xrft._stack_chunks(da2, dim).squeeze() if chunks_to_segments else da2
     da2_prime = da2_seg - da2_seg.mean(dim=dim)
     # Check that the (rectangular) integral of the cross-spectrum matches the windowed co-variance
-    npt.assert_almost_equal((1 / delta_xy) * cs.mean(fftdim).values, 
-                            ((da_prime*window) * (da2_prime*window)).mean(dim).values, 
+    npt.assert_almost_equal((1 / delta_xy) * cs.mean(fftdim).values,
+                            ((da_prime*window) * (da2_prime*window)).mean(dim).values,
                             decimal=5)
 
     ### Test Parseval's theorem for a 3D case with `window=True` and `detrend='linear'`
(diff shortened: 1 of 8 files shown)


navidcy:
@roxyboy, are there any tests for detrending? If no, we should add some :)

navidcy:
This PR attempts to resolve #111

navidcy:
My comment above didn't seem to result in the issue being linked to the PR... Could somebody link #111?

roxyboy:
@navidcy The test for detrending is currently here: https://github.com/xgcm/xrft/blob/8b7e84eb414ed02fff059998d2396430417a1d29/xrft/tests/test_xrft.py#L80

navidcy:
how can I run the tests locally to make sure they pass OK?

roxyboy:
`pytest xrft -v` should do the trick. You need to have `pytest` installed in your local enviroment.

navidcy:
OK, @roxyboy allow me to have a look. I'll ping you.

paigem:
Thanks @roxyboy for the instructions on running tests locally. 

After running the test on our polyfit update, I am getting an error. I believe the error may be because polyfit accepts an xarray array, whereas the previous scipy.signal.detrend expected a numpy array.

**Specifics of the error:**
The line in `test_xrft.py` that brings up the error is: `da_prime = func(da.data, axes=[2]).compute()`. Here, `da.data` (i.e. *not* an xarray array) is being sent in as an input to `func()`, and understandably yields the error `AttributeError: 'Array' object has no attribute 'polyfit'`. We have written it such that, for ffts over 1d, the xarray array `da` should be sent to `func()`.

We have rewritten `_apply_detrend()` such that all 1d ffts (whether chunked or not) will be detrended via polyfit. The tests are currently written according to the previous logic: if the data is chunked at all, run `func()`.

Looking at it now, I think the 1d polyfit implementation we added should probably lie outside of `func()` and probably just in `_apply_detrend()`, but the same line from `test_xrft.py` will still raise an error since it calls `func()` directly with a numpy array and axes of length 1.

**My main question is:**
Can I change `test_xrft.py`, if others agree that it makes sense to do so? I'll have to look more closely to determine how to best update the test if so, but I just wanted to know if this is an option.

@navidcy what do you think?

navidcy:
Yes, we should change the tests to actually test what's implemented ;) Or write an additional test if needed.

(sorry I didn't have time to have a look yet...)

paigem:
7b4eafd: In this commit I moved the 1d polyfit implementation before any of the multidimensional screening in `func()`. This was done to help pass the tests that Navid updated.

xrft/tests/test_xrft.py
```diff
@@ -80,7 +80,7 @@ def numpy_detrend(da):
 def test_detrend():
     N = 16
     x = np.arange(N+1)
-    y = np.arange(N-1)
+    y = np.arange(N-1) + 0.1*np.random.rand(N-1)
```
navidcy, line 83: I added some non-triviality in the y axis (that is being detrended in the 1D FFTW tests. Otherwise, tests were failing simply because of slightly different implementation of detrending done via `scipy.signal.detrend()` and ours via `xarray.polyfit()`. We are simply removing the first order polynomial fit while `scipy.signal.detrend` does something a bit more elaborate.

navidcy:
@roxyboy, what do you think of these changes me and @paigem made?

roxyboy:
Thanks for the edits @paigem and @navidcy !!
Travis-ci seems to have passed but simply not showing up here... https://travis-ci.org/github/xgcm/xrft
I can't seem to figure out where the code coverage is decreasing with the PR... Any ideas?

navidcy:
@roxyboy we modified the tests/code a bit; I think now it should be good for you to have a careful look ;)

navidcy:
`test_detrend()` now also creates a 3D dummy data array to test that calling detrend without any axes still does what it's supposed to do

rabernat:
Thanks so much for this awesome work @paigem and @navidcy!

I will try to review by the end of the week.

> My comment above didn't seem to result in the issue being linked to the PR... Could somebody link #111?

This needs to be done in the original comment by the primary author (in this case @paigem). I edited that post to add the link.

paigem:
Thanks @rabernat, and sorry @navidcy - I didn't realize that was on me to link the issue here. 

I will also note that I haven't addressed my second comment in Issue #111 where I suggested we raise a warning if the user enters anything other than 'constant', 'linear', or 'None' that a warning should pop up. I'll look into that now, but it probably fits in a different PR at this point. 

What's the correct etiquette here - should I raise a new issue with that same comment, and then we can close that issue when the associated PR is merged?

navidcy:
The #111 issue's title only points to it's "part 1". I suggest edit the issue to only refer to part 1 and submit a separate issue for the warning?

navidcy:
#111's title only points to part 1. Also all the discussion in the issue is about part 1 (except a remark by @roxyboy that "a warning is a good idea". May I suggest split the issue into two: create a new issue about the warning and ~~strikeout~~ the part two from #111?

paigem:
Sounds good @navidcy. I will edit issue #111 and create a new one for the warning.

roxyboy:
Thanks again @navidcy and @paigem . I think this PR is ready to merge.

xrft/tests/test_xrft.py
```diff
@@ -99,16 +110,25 @@ def test_detrend():
...
+    # test detrending along >1 dimensions
     da_prime = func(da.data, axes=[1,2,3]).compute()
     npt.assert_allclose(da_prime[0],
                         xrft.detrendn(d4d[0], axes=[0,1,2]))
 
+    da = da3d.chunk({'time': None})
+    # test detrending along all dimensions (total dimensions <4)
+    da_prime = func(da.data).compute()
+    npt.assert_allclose(da_prime,
+                        xrft.detrendn(d3d, axes=[0,1,2]))
```
rabernat, line 130: I am trying to understand how these tests work, and I'm getting tied up in circles. This is not the fault of this PR (tests were already there when you started), but I figure we should use this opportunity to try to make things a little more transparent.

What we have here is something like

```python
func = xrft.detrend_wrap(xrft.detrendn)
da3d = xr.DataArray(d3d, ...)
da = da3d.chunk({'time': None})
da_prime = func(da.data).compute()
npt.assert_allclose(da_prime, xrft.detrendn(d3d, axes=[0,1,2]))
```

It feels to me that this test is a bit circular. In the end, we are testing that the results of `detrendn` equal the results of `detrendn`!

Can we think of a way to refactor this test so that the purpose of each part is a little more clear and transparent?

If not, that is fine, and we can move forward without such changes.

xrft/xrft.py
```diff
@@ -116,25 +116,26 @@ def detrend_wrap(detrend_func):
...
         else:
             if len(set(axes)) < len(axes):
                 raise ValueError("Duplicate axes are not allowed.")
 
-        for each_axis in axes:
-            if len(a.chunks[each_axis]) != 1:
-                raise ValueError('The axis along the detrending is upon '
+            for each_axis in axes:
+                if len(a.chunks[each_axis]) != 1:
+                    raise ValueError('The axis that is being detrended '
```
rabernat, line 136: Is this still a limitation now that we are using xarray's  built-in `polyfit`? Can't we relax this constraint?

xrft/xrft.py
```diff
@@ -148,16 +149,15 @@ def func(a, axes=None):
 
 def _apply_detrend(da, axis_num):
     """Wrapper function for applying detrending"""
-    if da.chunks:
+    if len(axis_num) == 1:
+        func = detrend_wrap(detrendn)
+        da = func(da,axes=axis_num)
+    elif da.chunks:
```
rabernat, line 155: Do we really need this check and this wrapper now that we are using `polyfit`?

rabernat (commented):
Thanks a lot for this great PR!

In reviewing the PR, I felt a bit confused about all the layers of "wrappers" in the code. Some of these may be unnecessary now that we are using polyfit. It seems we have the opportunity to remove some code and simplify xrft, which is always a good thing.

We don't have to address these concerns right now. They are really more for @roxyboy, who is the architect of this feature. But I thought I would at least ask before we move forward.

xrft/tests/test_xrft.py
navidcy: Well... indeed @rabernat! They seem to be tautological. This was a problem with the test from before.
I made some changes (de512cf) to try to make the tests more strict actually asserting that the `detrend()` gives the expected output. This created a bit of boiler-plate code in `test_detrend()` -- I'll try to gather that up.

xrft/xrft.py
navidcy: hm... probably you are right @rabernat this may be redundant
our thinking with @paigem (if I recall correctly) was to deal with all detrending in the inner-most part of this wrapper... We (or at least I) were a bit confused with the wrappers.

@paigem, let's chat about how we can make this better or cleaner (or hopefully both)

xrft/xrft.py
navidcy: Well, I think it *is* if user asks to detrend *more than 1 dimensions*. In that case `xarray.polyfit()` cannot be invoked and we fall back to the previous `detrendn()` function.

@rabernat, the lines you are pointing to are only called when `len(axes)>1`; see https://github.com/xgcm/xrft/blob/78d90a21f6768330a79ef07657ed1e2810cda6e6/xrft/xrft.py#L122

Any suggestion how to make this cleaner?

navidcy: hm... I guess the question is whether we can use `xarray.polyfit` to somehow replace:
https://github.com/xgcm/xrft/blob/78d90a21f6768330a79ef07657ed1e2810cda6e6/xrft/xrft.py#L74-L102

to tell you the truth though I'm not 100% sure what the above-mentioned lines do... I guess somehow they best-linear hyperplane is computed and subtracted... I feel that there should be a place for `polyfit()` to do the job here, no?

navidcy:
perhaps it's a good opportunity to deal with #113 in this PR? (if #113 requires something to be dealt with that is)

xrft/xrft.py
paigem: Yes, @navidcy and I debated whether to put the polyfit detrending within the wrapper function or not. Happy to discuss this more @navidcy!

xrft/xrft.py
paigem: I agree that we should look into doing multidimensional detrending using `polyfit()`. I also find this bit of code confusing and thought we could make a first step by implementing `polyfit()` for 1d, and then figure out more dimensions. Let's look into this as well @navidcy!

rabernat:
> perhaps it's a good opportunity to deal with #113 in this PR? (if #113 requires something to be dealt with that is)

Yes, I would say, as long as your hands are deep into this code, this is a good chance to resolve #113.

xrft/xrft.py
```diff
@@ -116,25 +116,26 @@ def detrend_wrap(detrend_func):
...
             axes = tuple(range(a.ndim))
+
+        if len(axes) == 1:
+            poly_coefs = a.polyfit(dim=a.dims[axes[0]],deg=1)
+            return a - (poly_coefs.data_vars['polyfit_coefficients'][0]*a[a.dims[axes[0]]] + poly_coefs.data_vars['polyfit_coefficients'][1])
+
+        elif len(axes) > 3:
+            raise ValueError("Detrending is only supported up to "
+                            "3 dimensions.")
+
```
rabernat, line 129: I would recommend using the following functions here:

```python
def _detrend_dim(da, dim, deg=1):
    # detrend a single dimension
    p = da.polyfit(dim=dim, deg=deg)
    fit = xr.polyval(da.time, p.polyfit_coefficients)
    return da - fit

def _detrend(da, dims, deg=1):
    # detrend multiple dimensions in serial
    # only valid for linear case (deg=1)
    da_detrended = da
    for dim in dims:
        da_detrended = _detrend_dim(da, dim, deg=deg)
    return da_detrended
```

paigem: Thanks for your suggestion @rabernat. Using `polyval()` is clearly the way to go - didn't realize that was an option!

For the detrending along multiple dimensions, it looks like you are suggesting to loop through each dimension and perform the 1d detrend on each dimension. My understanding is that, if we want to do, for example, a 2d detrend (e.g. if I am calculating a FFT over 2 dimensions) we will need to fit a plane to the 2d data and subtract that off. I don't think this is the same as doing a 1d linear fit in the 2 dimensions, but perhaps I'm misunderstanding something here.

paigem: @navidcy and I briefly discussed trying to implement dask's `linalg.lstsq()` for more than 1d detrends, since I think `polyfit()` only works for 1d.

rabernat: > My understanding is that, if we want to do, for example, a 2d detrend (e.g. if I am calculating a FFT over 2 dimensions) we will need to fit a plane to the 2d data and subtract that off. I don't think this is the same as doing a 1d linear fit in the 2 dimensions, but perhaps I'm misunderstanding something here.

I think you're absolutely right. 😊  Now I am remembering we dealt with the same question when @roxyboy first implemented this feature. He actually implemented up to 3D detrending using fancy linear algebra (see [code here](https://github.com/xgcm/xrft/blob/8b7e84eb414ed02fff059998d2396430417a1d29/xrft/xrft.py#L53)). However, that code does not handle dask data properly in all cases.

Xarray's polyfit actually uses dask.linalg.lstsq under the hood. It is called via this sequence:
- https://github.com/pydata/xarray/blob/2ce1cfc3e33f4ac7b420a51ac54f39dff1ce008c/xarray/core/dataset.py#L6095-L6097
- https://github.com/pydata/xarray/blob/2ce1cfc3e33f4ac7b420a51ac54f39dff1ce008c/xarray/core/duck_array_ops.py#L624-L629
- https://github.com/pydata/xarray/blob/2ce1cfc3e33f4ac7b420a51ac54f39dff1ce008c/xarray/core/dask_array_ops.py#L115-L141

However, it may be very hard to do the 2D case this way from within xrft.

In order to keep this PR from getting too complicated, I recommend we do something like this:

```python 
def detrend(data, dims, ...):
    # data is still a DataArray at this point
    if len(dims)==1:
        return new_detrend_1D_with_polyfit(data, dims[0])
    else:
        return legacy_detrend_function(data, dims)
```

navidcy: I agree -- let's do this and open an issue for ≥2 -dimensional detrending

paigem: Sounds good to me!

navidcy:
@paigem, do lines 
https://github.com/xgcm/xrft/blob/4bdb6fce5b5e4bf96e7b579ac6f5c87ba3cca447/xrft/xrft.py#L130-#L132
needed or should I delete them completely?
(this is #113)

paigem:
Newest PR reorganizes the detrending portion of the code. With @navidcy. Also removed several lines that I deemed redundant, but I might have gone a bit overboard so best if others take a look as well. Major changes:

- both constant and linear detrends are now done in `apply_detrend()` (note that the leading underscore has been removed, since this function is now exported so it can be tested)
- we now use `polyval()` in conjunction with `polyfit()` per @rabernat's suggestion above
- added a line about valid detrend arguments to address #114 
- removed warning in `dft()` for axes being detrended that have chunks not equal to 1, since this appears to be addressed by simply defining chunks of length 1 (currently being done within `apply_detrend()`

paigem:
> @paigem, do lines
> https://github.com/xgcm/xrft/blob/4bdb6fce5b5e4bf96e7b579ac6f5c87ba3cca447/xrft/xrft.py#L130-#L132
> needed or should I delete them completely?
> (this is #113)

Good point - I had just looked at this! I personally think they can be deleted.

navidcy:
@paigem, what about
https://github.com/xgcm/xrft/blob/4bdb6fce5b5e4bf96e7b579ac6f5c87ba3cca447/xrft/xrft.py#L123-#L126
shall I delete them (see #113)?

paigem:
> shall I delete them (see #113)?

@navidcy I was just looking at this, and I think we can go ahead and delete these lines, assuming the tests still pass.

xrft/xrft.py
```diff
@@ -15,7 +15,7 @@
 import scipy.linalg as spl
 
 
-__all__ = ["detrendn", "detrend_wrap",
+__all__ = ["detrendn", "detrend_wrap","apply_detrend",
```
navidcy, line 18: One remark: the `apply_detrend` method was upgraded from private to public
This was mostly done so that it could be tested. Perhaps there is an alternative way to use a private function in tests other than upgrading it to public?

If we decide to keep `apply_detrend` as public then it deserves an upgrade in its docstring! :)

xrft/xrft.py
```diff
@@ -116,50 +116,52 @@ def detrend_wrap(detrend_func):
...
+            return da - fit
+
+        elif len(dim) > 3:
+            raise ValueError("Detrending is only supported up to "
+                            "3 dimensions.")
+
+        # For > 1d detrend, need all non-FFT axes to have chunks of length 1
+        for d in da.dims:
+            if d not in dim:
+                da = da.chunk({d:1})
```
rabernat, line 153:
```suggestion
                raise ValueError(f"Cannot apply detrending over dimension {d} because it is chunked. "
                                 f"Please rechunk your data first by calling `da.chunk({{d}: 1)`. ")
```

I would definitely not recommend silently and automatically rechunking the data. This could lead to big problems for large datasets. Instead, raise an error and ask the user to explicitly rechunk the data.

We would also need to make sure this scenario is covered by a test (e.g. `pytest.raises`).

xrft/xrft.py
rabernat: I didn't think any of the detrend functions should be part of the public API. Although I can see how the high-level `apply_detrend` could be a useful standalone utility.

Private functions should start with an underscore.

Regardless, you can still import private functions from other modules in the tests, e.g. `from xrft.xrft import _detrendn`.

rabernat (changes requested):
Thanks so much for continuing to iterate on this. I think we are really almost there!

navidcy:
I made `_apply_detrend` private and we can discuss whether `apply_detrend` and/or `apply_window` should be made private in a separate issue/PR

xrft/xrft.py
paigem: Sounds good. The current version of the code does this chunking, and to be honest I don't fully understand why it's needed (beyond the fact that it's for the 2d and 3d detrending via matrix algebra) so left it in. But I agree, we should raise an error instead of doing automatically.

paigem: Clarification: these lines here actually chunk the data into chunks of length 1, so the error should state that dimensions not being detrended over must be chunked. I deleted a warning in the last PR that I will add back in to this effect.

navidcy:
found a few cases that throw warning errors that were not covered in the tests and fixed
now codecov is happy :)

navidcy:
I'll stop.. I feel the `test_detrend()` is now sort of overkill :)

@rabernat, have a look at the changes we've made with @paigem at your convenience

rabernat (approved):
This looks great! Thanks for your patience digging so deep into this messy code. Really appreciate the hard work here!

navidcy:
@rabernat, perhaps a new release after this merge?

PR #117, opened 2020-10-30
[WIP] Make sure xarray.polyfit does not rechunk datarray

### LINKED ISSUE #116: Undesired chunking of 3d input when using linear detrend with `poyfit()`
I am running the most up-to-date version of xrft, that uses xarray's `polyfit()` for linear detrending. 

When I run `xrft.dft()` on one dimension with linear detrending on a 3d input, I am getting an output that has different chunking than the original input. Specifically, the first axis that is not the fft dimension appears to be output with smaller chunks.

For example, if I construct a simple 3d input:
Screen Shot 2020-10-30 at 11 38 11

Then the output is as follows. Note the different chunking along the 'lon' axis. 
Screen Shot 2020-10-30 at 11 39 39

In the above image, I have printed out the original input array chunks, the chunks that result from applying `polyfit()`, and the resulting chunks after all detrending. It thus appears that `polyfit()` is rechunking the 'lon' dimension from chunks of length 2 to chunks of length 1, and that this persists after the detrending function.

Any idea if this is normal behavior? I wonder if it might be a bug in xarray's `polyfit()` implementation, but perhaps I have misinterpreted how it is supposed to work. I'm currently testing whether rechunking after detrending to the original chunk sizes would be a workaround for xrft's `polyfit()` implementation.

This rechunking doesn't cause problems if the data is not very large, but when I try to run this on, e.g., the global CESM ocean data, dask can't handle it (presumably due to very small chunk sizes/large number of tasks).

Discussed this issue with @navidcy

navidcy:
Out of principle, `xrft` should not be rechunking the users input on the background. This PR is a "hack" for what supposed to be an issue with `xarray.polyfit()` (see #116 and https://github.com/pydata/xarray/issues/4554). `xarray.polyfit()` seems to be rechunking the data and `xrft` calls `xarray.polyfit()` when the user asks for a linear detrend of the input of `xrft.dft()`. The proposed change simply makes sure that the detrended data array has the same chunking as the original one.

Closes #116

### DIFF
--- a/xrft/xrft.py
+++ b/xrft/xrft.py
@@ -141,7 +141,8 @@ def _apply_detrend(da, dim, axis_num, detrend_type):
         if len(dim) == 1:
             p = da.polyfit(dim=dim[0], deg=1)
             linear_fit = xr.polyval(da[dim[0]], p.polyfit_coefficients)
-            return da - linear_fit
+            da_detrended = (da - linear_fit).chunk(da.chunks) # see discussion in https://github.com/xgcm/xrft/issues/116
+            return da_detrended
 
         elif len(dim) > 3:
             raise NotImplementedError("Detrending over more than 4 axes is "


xrft/xrft.py
```diff
@@ -141,7 +141,8 @@ def _apply_detrend(da, dim, axis_num, detrend_type):
         if len(dim) == 1:
             p = da.polyfit(dim=dim[0], deg=1)
             linear_fit = xr.polyval(da[dim[0]], p.polyfit_coefficients)
-            return da - linear_fit
+            da_detrended = (da - linear_fit).chunk(da.chunks) # see discussion in https://github.com/xgcm/xrft/issues/116
```
rabernat, line 144: This will not fix the problem. Instead this just creates a new layer of chunks in the task graph. The "upstream" tasks, which have the problems described in #116, still exist and have to be computed before we get to the new chunks. So actually this makes things worse by increasing the total number of tasks.

rabernat (commented):
Thanks for trying this @navidcy. However, I don't think the fix can possibly be so easy.

rabernat:
FYI, I am working on this a bit.

PR #118, opened 2020-10-30
Refactor detrend

rabernat:
This PR refactors detrending by moving the detrending function to a standalone module and testing it in isolation.

I also rewrote the detrending functions in a cleaner way. I did NOT yet implement 3D detrending. Do people think that 3D detrending is important?


xrft/detrend.py
```diff
@@ -24,9 +25,36 @@ def detrend(da, dim, detrend_type='constant'):
...
+    assert arr.ndim == 2
+    N = arr.shape
+
+    col0 = np.ones(N[0] * N[1])
+    col1 = np.repeat(np.arange(N[0]), N[1]) + 1
+    col2 = np.tile(np.arange(N[1]), N[0]) + 1
+    G = np.stack([col0, col1, col2]).transpose()
+
+    d_obs = np.reshape(arr, (N[0] * N[1], 1))
+    m_est = np.dot(np.dot(spl.inv(np.dot(G.T, G)), G.T), d_obs)
```
roxyboy, line 58: One way to simplify the code would be to estimate `m_est` using the psuedo inverse, i.e. `m_est = np.pinv(G).dot(d_obs)`. I think this would allow for better compatibility with `dask`.

roxyboy (commented):
As a thought, I think using the psuedo inverse to estimate the least-square fit could make the code simpler. Parallelizing `np.pinv` and `np.dot` is pretty straight forward.

xrft/detrend.py
rabernat: This is a good point. However, dask's fft functions (which we rely on) all require that the array be contiguous over the FFT dimensions anyway. So we just want to apply whatever detrending function in parallel across many chunks. apply_func gets the job done.

rabernat:
This is now ready for review. @navidcy and @paigem--would be great if you could have a look!

navidcy:
👍 
@rabernat, add me as a reviewer

rabernat:
You are not a maintainer on this repo (although it would make sense to add you), so I can't tag you as a reviewer. But that doesn't stop you from reviewing!

navidcy:
gotcha -- sure, I will review regardless. hopefully later today..

(also happy to be added as maintainer)

xrft/detrend.py
```diff
@@ -0,0 +1,71 @@
...
```
navidcy, line 68: 1. When we were first looking with @paigem these lines we were puzzled.
Now, after some fiddling I understood what's happening here. But it wasn't trivial.
Given that the test pass I'm optimistic that the transposes and the inversions are done correctly. However, should we add few comments or, perhaps, a reference to somewhere to ease the life of someone who would potentially trying to understand what's happening here in the future?

2. Would it be clearer if for constructing `col1`, `col2` we first do something like
```python
dim1, dim2 = np.meshgrid(np.arange(N[0])+1, np.arange(N[1]+1)
```
and then converted the arrays `dim1` and `dim2` in columns?

xrft/tests/test_detrend.py
```diff
@@ -0,0 +1,106 @@
...
+    assert detrended.chunks == da_trend.chunks
+    if detrend_type is None:
+        xrt.assert_equal(detrended, da_trend)
+    elif detrend_type == "constant":
+        xrt.assert_allclose(detrended, da_trend - da_trend.mean(dim=detrend_dim))
+    elif detrend_type == "linear":
+        xrt.assert_allclose(detrended, da_original)
+
+
+# always detrend on x y dims
```
navidcy, line 75: I'd add another comment above in the spirit of `# following tests 2D detrending`

xrft/tests/test_detrend.py
```diff
@@ -0,0 +1,106 @@
...
+    assert len(dims) == len(shape)
+    coords = {d: (d, np.arange(n)) for d, n in zip(dims, shape)}
+    data = np.random.rand(*shape)
+    for n in range(len(shape)):
+        data = sps.detrend(data, n)
+    da = xr.DataArray(data, dims=dims, coords=coords)
+    return da
+
+
+@pytest.mark.parametrize(
```
navidcy, line 27: I'd add another comment above in the spirit of `# following tests 1D detrending`

xrft/detrend.py
```diff
@@ -0,0 +1,71 @@
...
+import xarray as xr
+import scipy.signal as sps
+import scipy.linalg as spl
+
+
+def detrend(da, dim, detrend_type="constant"):
+    """Detrend a DataArray along dimensions described by dim"""
+
+    if detrend_type not in ["constant", "linear", None]:
+        raise NotImplementedError(
```
navidcy, line 15: does this needs to be tested?

navidcy (approved):
I made a remark on the human-interpretability of the 2D detrending algorithm. It'll be good if we can clarify a bit what's happening there...

Otherwise, this looks good. And it's much cleaner than before!

navidcy:
@paigem perhaps you can put this PR to the "ultimate" test and see if it can handle detrending your ESM daily data?

navidcy:
> Do people think that 3D detrending is important?

I don't think it's that important. If this is merged with the 3D detrending missing we should just add an issue mentioning this.

paigem:
Thanks @rabernat for this update! Somehow I missed the initial comment from @rabernat's update and so am only just seeing all this activity now! Will try to review it later today.

xrft/detrend.py
```diff
@@ -0,0 +1,71 @@
...
+    elif detrend_type == "constant":
+        return da - da.mean(dim=dim)
+    elif detrend_type == "linear":
+        data = da.data
+        axis_num = [da.get_axis_num(d) for d in dim]
+        chunks = getattr(data, "chunks", None)
+        if chunks:
+            axis_chunks = [data.chunks[a] for a in axis_num]
+            if not all([len(ac) == 1 for ac in axis_chunks]):
+                raise ValueError("Contiguous chunks required for detrending.")
```
paigem, line 31: By "contiguous chunks", do you mean that the dimensions being detrended over cannot be chunked?

xrft/xrft.py
```diff
@@ -375,30 +239,17 @@ def dft(
...
-            dim = [
-                dim,
-            ]
+            dim = [dim]
     if real is not None and real not in dim:
         dim += [real]
 
-    if not da.chunks:
-        if np.isnan(da.values).any():
-            raise ValueError("Data cannot take Nans")
```
paigem, line 400: Why is this check for NaNs not necessary anymore? I don't believe that `sps.detrend()` works on NaNs. 

If NaNs might be problem, is it worth having a test to verify that NaNs won't cause a problem for the detrending?

paigem:
> @paigem perhaps you can put this PR to the "ultimate" test and see if it can handle detrending your ESM daily data?

Great suggestion @navidcy! I just tried it, and this new PR works great on the CESM data! I just had to use `.fillna(0)` to fill the NaNs in the dataset with zeros before running xrft's `dft()`.

xrft/xrft.py
paigem: After a bit more testing, I see that `scipy.signal's detrend()` already has such an error for NaNs, so perhaps it does make sense to not duplicate here. If it's generally useful, we could add a comment about using `.fillna(0)` to fill NaNs with zero. But, as I mentioned in my comment below, this may not make sense in all situations.

navidcy:
I'm not sure if filling with zeros is a good idea...

If NaNs are there because you are looking e.g. at the SST over land then it's OK.
But if NaNs are there simply because your time-series misses some values for one reason or the other then it's not OK..

```python
import numpy as np
import scipy.signal as sps
import matplotlib.pyplot as plt
plt.close("all")
n = 20

noise = np.random.randn(n)
noise = sps.detrend(noise)

trend = 0.2*np.arange(n)
signal1 = trend + noise

plt.figure(figsize=(8, 8))
plt.subplot(221)
plt.plot(trend, label="trend")
plt.plot(signal1, marker="*", label="signal")
plt.title("signal")
plt.legend()

plt.subplot(222)
plt.plot(noise, marker="o", label="original noise")
plt.plot(sps.detrend(signal1), marker="x", label="detrended signal")
plt.title("detrending")
plt.legend()

trend = 0.2*np.arange(n)
signal2 = trend + noise
signal2[5:12] = np.NaN

signal2_filledzeros = signal2.copy()
signal2_filledzeros = np.where(np.isnan(signal2), 0, signal2)

plt.subplot(223)
plt.plot(trend, label="trend")
plt.plot(signal2, marker="*", label="signal")
plt.title("signal with some NaNs")
plt.legend()

plt.subplot(224)
plt.plot(noise, marker="o", label="original noise")
plt.plot(sps.detrend(signal2_filledzeros), marker="x", label="detrended signal after filling 0")
plt.title("detrending after filling with 0")
plt.legend()
```

mwe

paigem:
Thanks for the nice demonstration of filling with zeros @navidcy! Clearly filling with zeros does not always make sense, and so we certainly shouldn't implement something of the sort in `xrft()`.

Moving forward with this PR, I think we need to decide if we want to support NaN values or not. I see three options:
- Leave the PR as is, since `scipy.signal.detrend()` throws an error when there are NaNs. This will force the user to think about how to deal with NaNs (as it did for me!). I think this would be just fine.
- Implement different code that allows for NaNs in the data. I believe that xarray `polyfit()` does this.
- Leave the PR mostly as is, but add a specific error about NaN values, with a suggestion that `.fillna()` might be one avenue, particularly to fill land points when detrending over time.

I vote for the first or third options in the short term, and the second in the long term. Or maybe @rabernat or @roxyboy have some more sophisticated ideas here! 🙂

xrft/detrend.py
rabernat: Yes, that's what I meant. I will update to be more clear.

xrft/detrend.py
rabernat: Good suggestion. I'll implement this.

xrft/detrend.py
rabernat: Yes it should, good catch.

rabernat:
Thanks for the helpful and constructive reviews.

Here are my thoughts on NaNs. It's very important to recognize two linked points:
- In xrft, we only ever detrend over dimensions that we are also taking FFTs over
- FFT (as implemented in numpy / dask) does not work in the presence of NaNs

So, if you have any NaNs in your data, xrft will definitely not work in those locations.

My preference would be to **document this very clearly** but **not explicitly check for NaNs**. Rather, I'd prefer to let the upstream functions (fft, sps.signal.detrend) do their thing. The reasons for this are:
- It keeps xrft simple, reducing the amount of code we have to maintain
- Pre-checking for NaNs can be expensive for large data, and supporting large data is an explicit goal of xrft

If you agree with this, I think the only change needed in this PR is to update the documentation to clarify the treatment of NaNs.

paigem:
> Here are my thoughts on NaNs. It's very important to recognize two linked points:
> - In xrft, we only ever detrend over dimensions that we are also taking FFTs over
> - FFT (as implemented in numpy / dask) does not work in the presence of NaNs

In response to your first point here @rabernat, `scipy.signal.detrend()` seems to raise an error if there are any NaNs in the dataset, not just in the dimension being detrended over. So I agree with your above statements, but just want to make sure we are aware that any NaNs in a dataset will raise an error if doing a linear detrend.

Still, I agree that we should not handle NaNs ourselves and should instead include a detailed write-up in the documentation about the use of NaNs.

paigem (approved):

roxyboy:
Considering that the FFT itself won't work with data having NaNs and not just for detrending, I think documenting it would be the way to go.

paigem:
Hmm, @roxyboy I seem to be able to run `xrft.dft()` successfully with NaNs in my data if `detrend=None` or `detrend='constant'`. I only get an error raised if I have `detrend='linear'` (and the error is from scipy.signal).

roxyboy:
@paigem ...that seems kind of weird. Do your `dft` outputs not include nans?

paigem:
@roxyboy yes, the output also contains nans. It looks like any nan is propagated through the detrending dimension. That is, if we have, e.g., a 2-d dataset in (time,lon) with a nan at [1,1] and want to take an fft in time, then no fft is taken at point lon=1. This behavior is exactly what I would want when the only nans are land points. 

Here is a screenshot of my test case that shows this behavior:
Screen Shot 2020-11-13 at 12 08 26
Screen Shot 2020-11-13 at 12 08 53

So anyway, it looks like nans won't raise an error except in the case of linear detrending. I still don't think this alters the current PR, but should probably be included in the documentation write-up.

roxyboy:
Since your first row contains a NaN (`A[1,1] = np.nan`), the resulting FFT gives all NaNs in the corresponding longitude. An FFT that returns all NaNs is expected when there are any NaNs in the original field and I wouldn't call this as being able run successfully... The user will realize that the output has NaNs and will have to do something about in their preprocessing of data.

paigem:
Sounds good @roxyboy!

PR #120, opened 2020-11-02
CI migration to gh actions + Black formatting

jbusecke:
This PR moves the CI to github actions and implements linting with black as part of the CI.
- [x] Closes #115

PR #122, opened 2020-11-10
Phase corrections

lanougue:
Forward transform has been update to match with theoretical Fourier Transform.

1) Coordinates range is carrefully taken into account.
2) An fftshift is applied before fourier transform. This is done in order to anticipate that numpy.fft implicitely assumes that input signal is on a fftshifted grid.


lanougue:
Here is a descriptive of why these changes are desirable in dft.
(I apologize for the huge text police in the pdf file...).
[forward_phase_correction.pdf](https://github.com/xgcm/xrft/files/5519425/forward_phase_correction.pdf)

xrft/xrft.py
```diff
@@ -426,23 +431,37 @@ def dft(
...
 
     newdims, newcoords = _new_dims_and_coords(da, axis_num, dim, k, prefix)
+    
+    print(newdims)
+    print(newcoords)
+    updated_dims = [newdims[i] for i in da.get_axis_num(dim)]
+    print(updated_dims)
+    phase = 1.
+    for up_dim, l in zip (updated_dims, lag):
+        phase = phase*xr.DataArray(np.exp(-1j*2.*np.pi*newcoords[up_dim]*l), dims=up_dim, coords={up_dim:newcoords[up_dim]}) # taking advantage of xarray automatic broacasting and ordered coordinates
```
roxyboy, line 458: Some of the tests seem to be failing due to this line.

roxyboy (commented):
This PR looks good, thanks @lanougue ! The tests seem to be failing due to a data type incompatibility: `TypeError: unsupported operand type(s) for *: 'complex' and 'cftime._cftime.DatetimeJulian'` Do you think you can make it compatible with `cftime`?

xrft/xrft.py
lanougue: Ok, I hope it works now.

roxyboy:
Thanks for the edits @lanougue ! 
The tests are now all passing except for one with an error message: `Error: The process '/opt/hostedtoolcache/Python/3.8.6/x64/bin/pre-commit' failed with exit code 1`. I've never encountered such error but does @rabernat or @jbusecke know where this is coming from...?

jbusecke:
That's the new code Formatting CI I have implemented in #120. The "contributors guide" in the docs describes how to satisfy this one. I however had some typos in there so rather take a look at #123. Let me know if that is still unclear!

lanougue:
I ran black on the code. It has reformatted stuff in the modifications I did. But it still does not pass the test. Could the error be more verbose in order to understand what is goind on ?

roxyboy:
Awesome, the tests seem to have all passed!! @rabernat Do you think we can merge this PR??

roxyboy:
@lanougue As a minor final suggestion, do you think you could add a test that examines the phase? An example would be where you compare the `dft` and theoretical transforms as you already show in your PDF. I think this would highlight the significance of your edits.

lanougue:
@roxyboy , ok, I will try to implement a test with the theoretical result I used in the PDF

lanougue:
I added a test that checks if the phase is correctly preserved. It works as follow:  two identical signals are generated (same amplitude at the same coordinate).  On both signals, different random number of zeros are added on each sides. Then, their two dft are verified to be the same.

This test only checks for phase preservation. It does not check consistency with theoretical Fourier Transform. Testing that last point should be discussed (if feasible...)

roxyboy:
@lanougue Maybe I'm misunderstanding something but couldn't we simply add `numpy.assert_almostequal(xrft.dft(s2).real*dx, TF_s.real)` and `numpy.assert_almostequal(xrft.dft(s2).imag*dx, TF_s.imag)` as you have at the end of your PDF example now that `xrft` takes care of the phasing?

lanougue:
Hi @roxyboy. Yes, we could do as you are suggesting but the check would be limited to this simple example. Taking care of the correct phase is
1) verifying that translating the coords support does not affect the dft (1st test already implemented)
2) An fftshift is applied prior passing to fft.  Does it have to be tested ? In my opinion, it's kind of basic code that sould not be changed

roxyboy:
@lanougue  Perhaps we can add your PDF example in the documentation but actually seeing the results of `dft` and theory matching was convincing for me... Tests should be as simple as possible so I actually think your PDF example would be great for a test.

lanougue:
@roxyboy , @rabernat . I was thinking of some phase effects that could be undesirable for the users. For example, when the user compute cross-spectrum between two different signals (a functionality of xrft), having phase preserved in dft will result in: 1) the actual cross-spectrum phase but also 2) an additional phase that could come from the different coordinates support (lag between the signals coords range). The sum of this two phases is the true phase of cross-spectrum but could be undesirable if the user only cares about the first part (which was the default output at time)..
If we force the user to have preserved phase, he would have to mannually remove/change the coords of its two signals before computing the cross-spectrum. Something which could be annoying.

Thus, I would suggest to set a new argument in dft: "ignore_coords" which would be set to True by default. It would ensure a gentle transition between xrft versions (no changement in dft default behaviour mimicing np.fft) but will allow the user to carrefully take care of the coordinates if required.
Maybe, in the future, this argument could be passed to False by default depending on users feedback.
What do you think ?

lanougue:
@roxyboy , ok for the theory test, I will work on that

roxyboy:
@lanougue I like your idea of the `ignore_coords` flag. We would need to explain carefully what it does in the docstring of `dft`.

lanougue:
I think "ignore_phase" even more confusing. There will always be a phase, even a wrong one. Ignoring phase would say, for me, that only magnitude would be considered.

roxyboy:
Maybe `preserve_phase = False` could be the default to give the same behaviour as `np.fft`?

lanougue:
Yes, I like "preserve_phase" with default to False. I would just rename it
1. preserve_absolute_phase
2. preserve_full_phase
3. preserve_entire_phase
4. preserve_total_phase (my prefered)
What is your preference ?

roxyboy:
I'm fine with `preserve_total_phase`.

lanougue:
"preserve_true_phase" or just "true_phase" is even better, don't you think ?

roxyboy:
I feel putting `true_phase` would confuse the users who are just interested in the amplitudes as in the spectrum...

lanougue:
I think that users who only care about spectrum and amplitude won't see the difference since phase will vanish in both cases.
I definitely vote for true_phase ! :)
Maybe @rabernat could say.

xrft/xrft.py
```diff
@@ -462,18 +461,16 @@ def dft(
             )
         delta_x.append(delta)
         lag_x.append(lag)
-        shift_axes.append(d)
+    print(lag_x)
```
roxyboy, line 464: Should we make `xrft` print the lag?

roxyboy (commented):
Is there benefit in making `xrft` print the lag?

xrft/xrft.py
lanougue: no...my mistake

lanougue:
Now, it should be all right. It works for uncentered coords and odd number of points (ifftshift VS fftshift stuff) and matches theoretical results in all cases.
I have also written code for idft which follows the same conventions and works like a charm (signal is well recovered when phase is preserved).
As soon as this PR is accepted, I will propose a new PR for the addition of idft.
Now, it's up to you ! :)

xrft/xrft.py
```diff
@@ -417,23 +449,29 @@ def dft(
...
 
     if detrend:
         da = _apply_detrend(da, dim, axis_num, detrend)
 
     if window:
         da = _apply_window(da, dim)
 
-    f = fft_fn(da.data, axes=axis_num)
+    if true_phase:
+        f = fft_fn(np.fft.ifftshift(da.data, axes=axis_num), axes=axis_num)
```
rabernat, line 472: What happens here when `da.data` is a dask array? Is it computed eagerly?

rabernat (commented):
This is fantastic! Thank you some much for your contribution.

My one concern is how this new feature interacts with dask data. My guess is that it will coerce dask arrays to numpy. Have you checked this? We will need a test case to ensure that xrft always operates lazily.

xrft/xrft.py
rabernat: Fixing this is probably as simple as:
```suggestion
        f = fft_fn(fft.ifftshift(da.data, axes=axis_num), axes=axis_num)
```

lanougue: Thanks @rabernat !
I agree whith your suggestion (corrected in the new commit). I have done a small check on how it behaves with dask array and it seems to be ok (lazy computation). Is it everything good now ?

roxyboy: @lanougue Could you also update `test_phase_preservation` to test for dask?

lanougue: @roxyboy : I don't understand what I should do on this test function. What I understand from @rabernat comment is that it should exist another test that checks if dft is actually computed lazily with dask arrays. Am I right ?
I do not beleive that it should be tested in test_true_phase_preservation().

Should we do another test that checks for lazy computations with dask arrays? If yes, I propose testing dtype of dft output when passing dask arrays as input.

roxyboy: @lanougue Ok, considering that the test data in `test_phase_preservation` is one-dimensional, I guess it doesn't make sense to test `dask` there...

rabernat:
Hi @lanougue -- I just pushed an update to your branch that shows how to parametrize the test for dask.

Looking more closely at your test, it looks like this test does not actually examine and verify the coordinates on the outputs of `dft` with `true_phase=True`. You only check the values of the results, not the coordinates.

The recommended way to do this would be to create a DataArray with the expected dimensions and coordinates and the use xarray's testing to assert the output DataArray (data AND coordinates) matches the expected.

xrft/xrft.py
rabernat: I just addressed this in my latest push to @lanougue's branch.

We need to make sure that we explicitly check that computation remains lazy when calling `dft` with `check_phase=True`. It doesn't matter if Dask is realistically needed / reasonable in this particular test case. The laziness is the important property.

lanougue:
Hi @rabernat and thanks for the update. What I understand about this testing philosophy is that I should create a deterministic example for which I KNOW the output. Then I hardly code this xarray in the test (or find a way to compute it) and compare it with dft output. In case dft output is modified in the future, this test will check is the phase continue to be preserved. Am I right ?
If so, I can modify the test to do so.

roxyboy:
@lanougue I believe the procedure you described is right. You probably already know this (but just as a caveat as I myself have messed up git many times), you should probably `git fetch` and `rebase` @rabernat 's pushed branch and then add your edits to that branch and then push that again.

lanougue:
Oups, I just did "git pull origin phase_corrections". I am still learning git..
Modifications of @rabernat seems to have been taken into account.
What is the difference between this and fetch + rebase ?

rabernat:
> What I understand about this testing philosophy is that I should create a deterministic example for which I KNOW the output. Then I hardly code this xarray in the test (or find a way to compute it) and compare it with dft output.

Yes this is exactly right. IIUC, that is what your existing test does for the values. But since `true_phase` also involves coordinates, it is important to explicitly test the coordinates as well. So you just need to update your test to also check that the coordinates are as expected.

roxyboy:
@lanougue Based on what you described, it seems that you're making edits to your master branch, which is probably fine as you know already that your code works. So in this case, `git pull` should work just fine. But if you want to test out edits for example, I'd recommend making the edits in the future on a new [branch](https://git-scm.com/docs/git-checkout) and push that branch to `origin`. You generally don't want to edit your master branch as that can mess up the git status pretty quickly.

lanougue:
@rabernat 
To comply  with your comment, I suppose we can just change the last line of test_true_phase_preservation() by `xrt.assert_allclose(S1,S2)`. Is that ok ?
As it, this test will check the phase preservation by changing the coordinates of the input while not changing the signal amplitude and support.

Now, to verify that we actually get the expected output I suppose I can make another test.
As suggested by @roxyboy (in an earlier comment) this new test would be mixed with the theoretical Fourier Transform matching.

roxyboy:
@lanougue numpy.testing won't assert the coordinates with `xrt.assert_allclose(S1,S2)`. In addition to `xrt.assert_allclose(S1.values,S2.values)`, you also want to assert `xrt.assert_allclose(S1.coords, S2.coords)`

roxyboy:
Actually, `xarray.testing` also seems to test the coordinates so scrap what I just said...

rabernat:
> I suppose we can just change the last line of test_true_phase_preservation() by `xrt.assert_allclose(S1,S2)`. Is that ok ?

Yes this sounds right.

> Now, to verify that we actually get the expected output I suppose I can make another test.

Since `true_phase` only affects the coordinates, I'm not sure this is necessary. We already have good tests for dft in general. But your suggestions are welcome here. The more comprehensive and thorough the tests, the better.

lanougue:
@rabernat. I don't understand your comment. true_phase option do not affect the coordinates. It uses coordinates of the input to evaluate the true phase of the dft output. Having true_phase = True or False do not change the output coordinates but changes the output values.
Talking about this, I think this PR passes all the tests because default behaviour of dft is with true_phase=False. For example, using true_phase = True as defualt would certainly make some test to fail. Especially the one when dft and np.fft values are compared.

rabernat:
> true_phase option do not affect the coordinates.

Yes I see now that I was incorrect. Sorry for rushing through this review...

I think what confused me was the addition of the new `_lag_coord` function. This PR is somehow changing how we set coordinates on output data, not just for `true_phase=True` but for all outputs. So that is the part that needs to be tested.

lanougue:
_lag_coords just evaluate the mean lag of the input coords. It does not change the output coords. "lag" variable is only used to correct the dft. Coordinates output stay untouched in this PR

lanougue:
To be more clear:
1- When true_phase = False: This is the traditional old behaviour. Coords and complex amplitude of dft output are as they always were. lag variable is evaluated but not used at all.
2- When true_phase = True: Output coords are the same as they always were (same as true_phase=False). Only complex amplitude of dft output is changed

rabernat:
Hi @lanougue--thanks for clarifying and sorry for being slow to understand. What you say is now obvious from reading the code more carefully.

lanougue:
Ok, I added one more test that checks dft output when true_phase is set to True.

@roxyboy  : I think that checking dft output against theoretical expression is not feasible and not desirable. FFT will never be strictly equivalent to a Fourier Transform. It would be converging to FT only if sampling is infinitely small. Thus, I think we should restrict us to numerical to numerical comparisons.

roxyboy:
@lanougue you can always loosen up the test by the `rtol` flag. I would prefer we keep the theoretical expression and loosen the test criteria.

lanougue:
Checks fails in that last commit do not seem to be linked to my update. Looking to details shows strange errors. @rabernat, @roxyboy , Can you have a look ?

lanougue:
@roxyboy ; ok, I will give a try. I think I can do a recurcive test that checks if it converges to the theoretical expression

rabernat:
I fixed the CI.

lanougue:
Here we are, I followed @roxyboy advice to make a test that checks the matching of dft with theoretical expression. As suggested, I had to decrease the absolute tolerance of xrt.assert_allclose() up to 1.e-3 to assert the matching. If we consider that this tolerance should have a lower value, we need to set a smaller spatial sampling (dx) to assert the matching again (which is the expected behaviour).
I think that atol = 1.e-3 is fine for this test.

roxyboy:
Cool, I think we're ready to merge this PR! Thanks @lanougue and @rabernat for the edits!

rabernat:
Ok, fantastic. We are done here. Thanks for you patience as we got the tests rights.

lanougue:
Yes ! Thanks to both of you !

PR #126, opened 2020-11-19
Moving spacing as coords attribute + some update

lanougue:
spacing information is currently added as a new coordinate into dft output. I believe it is a strange behaviour.
In this PR, I move frequency spacing information as an attribute of the coordinate itself.
It allows to recall spacing information with:
S = xrft.dft(..., dim='x')
print(S.freq_x.spacing)

Stored as a coordinate attribute, spacing information is kept while doing some operation on xarray S. Moreover, there is now a unique name "spacing".

There was also a bug in _new_dims_and_coords() when called on xarrays with dimensions without coordinates. I propose a simpler version of _new_dims_and_coords() without this bug.


lanougue:
@roxyboy , @rabernat . If you are ok with that, I can start updating the tests who check spacing variable

roxyboy:
@lanougue if you want to check if the tests work locally, you can also always run `pytest xrft -v` in the top directory of `xrft`.

lanougue:
> @lanougue if you want to check if the tests work locally, you can also always run `pytest xrft -v` in the top directory of `xrft`.

Thanks ! I did not know about that ;)

lanougue:
To summerize:

1) spacing (currently assign as a new coordinate) are now defined as an attribute  of the corresponding coordinate. The spacing of all transformed dimensions are all named "spacing" but they can be rercovered through their respective coordinate : output.freq_x.spacing is the spacing of freq_x

2) No more bug when calling transform on dimension without coordinate. (Strange because I have seen a test somewhere to check that). It rresolve bug1 of #127 

3) _new_dims_and_coords() function simplified

lanougue:
This last commit solves bug2 of #127. No more problem when calling the real transform whose dim is not listed as last argument of dim list.
I added simplifications of transpose operation without defining again the new name of transformed coordinates. It is now automatic.
Sorry for adding modifications in this PR which are not related to "moving spacing attribues"...

xrft/xrft.py
```diff
@@ -96,20 +96,20 @@ def _stack_chunks(da, dim, suffix="_segment"):
...
+# raise ValueError(
+# "The dimension along real FT is taken must "
+# "be one of the existing dimensions."
+# )
+# elif real != transdim[-1]:
+# transdim.remove(real)
+# transdim += [real]
+# da = da.transpose(*transdim)
+# trans = True
+# return da, trans
```
rabernat, line 112: No commented-out code please. Just remove the function if it is unnecessary. Git keeps the entire code history.

xrft/xrft.py
```diff
@@ -272,15 +253,31 @@ def dft(
...
     else:
         if isinstance(dim, str):
             dim = [dim]
-    if real is not None and real not in dim:
-        dim += [real]
+
+    if real is not None:
+        if real not in da.dims:
+            raise ValueError(
+                "The dimension along real FT is taken must be one of the existing dimensions."
```
rabernat, line 265:
```suggestion
                "The dimension along which real FT is taken must be one of the existing dimensions."
```

xrft/tests/test_xrft.py
```diff
@@ -969,7 +969,7 @@ def test_true_phase():
     f = np.fft.fftfreq(len(x), dx)
     expected = np.fft.fft(np.fft.ifftshift(y)) * np.exp(-1j * 2.0 * np.pi * f * lag)
     expected = xr.DataArray(expected, dims="freq_x", coords={"freq_x": f})
-    expected = expected.assign_coords(freq_x_spacing=1 / (len(x) * dx))
+    # expected = expected.assign_coords(freq_x_spacing=1 / (len(x) * dx))
```
rabernat, line 972: No commented-out  code.

rabernat (changes requested):
This is great. Thanks for taking the time to clean up and simplify the code. Just some minor nit-picks.

lanougue:
Hi @rabernat, What do you think about these changes ?

rabernat:
Fantastic, thank you!

Issue #87, opened 2019-11-16
Issues with isotropic_powerspectrum: dask and freq_r axis


dhruvbalwada:
The isotropic_powerspectrum has a few problems:
a) The first wavenumber returned along the 'freq_r' dimension is a nan. Surely this should be zero or something like that. 
b) Problems when using with dask arrays. 
I get the error `TypeError: Must specify minlength argument in da.bincount`

Here is an example to reproduce it : 
```
da = xr.DataArray(np.random.rand(5,30,30),
                  dims=['time', 'y', 'x'],
                  coords={'time': np.arange(5), 'y': np.arange(30), 'x': np.arange(30)})
```

This works:  
`daisops = xrft.isotropic_powerspectrum(da, dim=['x','y'])`

But if I make da a dask array (even if it is not chunked along the dimensions along which fft is being done), then it doesn't work:
`daisops = xrft.isotropic_powerspectrum(da.chunk({'time':1}), dim=['x','y'])`

roxyboy:
The `isotropic_powerspectrum` has not been developed for parallelization so I'm not surprised it fails for `dask.array`s

roxyboy:
Linking issue #9 here.

rabernat:
`isotropic_powerspectrum` should be able to work lazily. Your error suggests this might be due to the recent addition of `bincount` to dask's built in algorithms.

@dhruvbalwada - as you know we have the SWOT proposal due this week. Once that's done I'll be able to spend some time on this. Bottom line, there is nothing special about isotropic powerspectrum that should preclude the use of dask.

roxyboy:
Oh, if dask has the API of bincount, parallelizing `isotropic_spectrum` should be quite straightforward :)

roxyboy:
PR #128 fixed this issue.

PR #128, opened 2020-11-25
Parallelize isotropize

roxyboy:
I implemented @shoyer 's amazing `numpy_groupies` hack to parallelize the isotropization of spectra. This addresses issues #9 and #87 .

### DIFF
--- a/ci/environment-py3.6.yml
+++ b/ci/environment-py3.6.yml
@@ -1,4 +1,7 @@
 name: test_env_xrft
+channels:
+  - conda-forge
+  - defaults
 dependencies:
   - python=3.6
   - xarray
--- a/ci/environment-py3.6.yml
+++ b/ci/environment-py3.6.yml
@@ -4,6 +4,7 @@ dependencies:
   - xarray
   - dask
   - numpy
+  - numpy_groupies
   - pytest
   - pip
   - black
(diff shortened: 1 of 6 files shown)


rabernat:
This is cool Takaya! We could also potentially use this in xhistogram

I will review this after the Thanksgiving holiday.

p.s. @shoyer is not the creator of numpy_groupies.

rabernat:
p.s. Isn't this going into xarray itself? Is it necessary to put it in xrft?

roxyboy:
> p.s. Isn't this going into xarray itself? Is it necessary to put it in xrft?

I think it eventually is going to but I checked the `xarray` issue log and it seems that there's still some issues for it to be implemented upstream. So I figured I'd just add it to `xrft` in the mean time.

roxyboy:
@rabernat Do you think we should keep this PR out until `numpy_groupies` gets implemented more upstream?

rabernat (approved):
This looks great! Do we have tests for this feature already? If so, I approve.

roxyboy:
> This looks great! Do we have tests for this feature already? If so, I approve.

The `test_isotropize` tests isotropization for both chunked and non-chunked data so I think the tests cover this feature.

PR #129, opened 2020-12-04
Adding idft

lanougue:
Adding new function idft to do inverse Fourier Transform


lanougue:
Here is an addition of the idft function. True_phase option is also available to do correct phase inversion. A new argument "lag" is also added to allow the user to have arbitrary output coordinates.

I can add tests. What kind of tests do you think are relevant ?
I think that the two tests below should be done:
- idft(dft(S)) == S
- idft(dft(S, true_phase=True), true_phase=True) == S

roxyboy:
> I think that the two tests below should be done:
> 
> * idft(dft(S)) == S
> * idft(dft(S, true_phase=True), true_phase=True) == S

I agree with this.

xrft/xrft.py
```diff
@@ -333,6 +334,186 @@ def dft(
     )  # Do nothing if da was not transposed
 
 
+def idft(
+    da,
```
roxyboy, line 338: @lanougue Can you change the docstring so that it explains `idft`? I think the input should be named as `daft` and output as `da` to be consistent with `dft`.

lanougue: Ok, I will wait for discussion #130 to go to a decision before continuing

roxyboy: Should we also have a `xrft.ifft` function...?

lanougue: I think so. It is in the plan too

lanougue:
I am almost done,
I still have a question on how to resolve some tests. I want to check what happen if a user try to do: ifft(fft(s))

If the user try idft(dft(s)) with true_phase and true_amplitude set to True for both transformations, he will recover the original signal "s". No problem.

However, ifft(fft(s)) will almost never be equal to s. This is because initial coordinates of "s" (and thus the phase) have been lost during the operation. Thus, coordinates of "s" and "ifft(fft(s))" will never match (it will match only if "s" coordinates are centered (meaning phase=0). However, s.data and ifft(fft(s)).data match as with numpy.
(shifting operation can yet alter this equality).

I think this is a totally normal behaviour, but can be disturbing for a user who always tough that ifft(fft(s))==s. It is true for amplitude but this is not the case anymore for the coordinates unless they are correctly taken into account during both transformations (as with dft() and true_phase flag).

In an xarray framework, the unique option to have idft(dft(s))==s for both amplitude and coordinates is to convince people to abandon fft to go to dft.

roxyboy:
> However, s.data and ifft(fft(s)).data match as with numpy.

Thanks for your work @lanougue ! I'm in favor of `s.data = ifft(fft(s)).data` as that is what one would also expect from numpy.

roxyboy:
@lanougue I think you can have `xrft.ifft` raise a warning about the coordinate information getting lost in the operation and recommend the usage of `xrft.idft`.

lanougue:
Done.
In all functions I change dft calls into fft calls to keep consistency with current xrft behaviour.
It could be interesting to let the user choose between fft or dft call in other functions such as power_spectrum, cross_spectrum, ...

xrft/xrft.py
```diff
@@ -179,6 +181,28 @@ def _calc_normalization_factor(da, axis_num, chunks_to_segments):
         return [da.shape[n] for n in axis_num]
 
 
+def fft(da, **kwargs):
+    if kwargs.pop("true_phase", False):
+        print("true_phase argument is ignored in xrft.fft")
```
roxyboy, line 189: I generally think making functions print messages is not standard. Could you change these to warnings?

xrft/xrft.py
```diff
@@ -393,7 +429,10 @@ def idft(
     daft : `xarray.DataArray`
```
roxyboy, line 443: Can we change the output from daft to da and the input to daft?

roxyboy (commented):
@lanougue I think these changes are great! My suggestions are only minor. Also could you add the testing of Parseval's identity with the amplitude and phase being set to True? There's already a test function (`test_parseval`) that checks for it using `xrft.power_spectrum` so you could probably add it there.

xrft/xrft.py
```diff
@@ -267,7 +274,7 @@ def dft(
     """
 
     if not true_phase and not true_amplitude:
-        msg = "xrft.dft default behaviour will be modified in future versions of xrft. Use xrft.fft to ensure future compatibility and deactivate this warning"
+        msg = "xrft.dft default behaviour will be modified in future versions of xrft. Prefer xrft.fft to ensure future compatibility and deactivate this warning. Consider using xrft.dft for accurate coordinates phasing and FT amplitude handling."
```
roxyboy, line 277: Maybe we could phrase this like: "The flags `true_phase` and `true_amplitude` will be set to True in future versions of xrft to preserve the theoretical phasing and amplitude of FT. Consider using xrft.fft to ensure future compatibility with numpy.fft like behavior and to deactivate this warning."

xrft/xrft.py
```diff
@@ -192,12 +195,16 @@ def fft(da, **kwargs):
...
+    """
+    See xrft.idft for argument list
+    """
     if kwargs.pop("true_phase", False):
         print("true_phase argument is ignored in xrft.ifft")
     if kwargs.pop("true_amplitude", False):
         print("true_amplitude argument is ignored in xrft.ifft")
     if kwargs.pop("lag", False):
         print("lag argument is ignored in xrft.ifft")
+    msg = "xrft.ifft do not guaranty output coordinate phasing. Prefer xrft.dft and xrft.idft as forward and backward Fourier Transforms with true_phase flag set to True for accurate coordinates handling."
```
roxyboy, line 207: A small typo: does not guarantee correct coordinate phasing for its output. We recommend... with the true_phase flag set to True... coordinate handling.

roxyboy:
> It could be interesting to let the user choose between fft or dft call in other functions such as power_spectrum, cross_spectrum, ...

I agree with this. Could we just add the true_phase and true_amplitude flags to power_spectrum and cross_spectrum by also keeping the default as False?

xrft/tests/test_xrft.py
```diff
@@ -1033,3 +1033,53 @@ def test_idft_dft():
         FTs, shift=True, true_phase=True, true_amplitude=True, lag=mean_lag
     )
     xrt.assert_allclose(s, IFTs)
+
```
roxyboy, line 1036: Thanks for adding the tests. Can you put these to the test that already existed as `test_parseval`?

roxyboy (changes requested):

lanougue:
Hi @roxyboy ,
Thanks for your review. I think I did all the changes you have suggested.
I recreated two very small tests (1D and 2D) for the parseval identity checking with dft and true flags.
I did not use existing test_parseval function because this later tests several other things at the same time (windowing, ...)

Adding true_phase and true_amplitude flags in power_spectrum and cross_spectrum is feasible (fft call should be transformed into dft call). However, output phase and amplitude will of course be modified and we should evaluate the impact of these modifications in these functions (which is probably also desirable too).

xrft/tests/test_xrft.py
roxyboy: @lanougue It starts from line 548 of test_xrft.py.

xrft/tests/test_xrft.py
```diff
@@ -642,6 +642,50 @@ def test_parseval(chunks_to_segments):
...
+            "x": dx
+            * (
+                np.arange(-Nx // 2, -Nx // 2 + Nx)
+                + np.random.randint(-Nx // 2, Nx // 2)
+            )
+        },
+    )
+    FTs = xrft.dft(s, dim="x", true_phase=True, true_amplitude=True)
+    npt.assert_almost_equal(
+        (np.abs(s) ** 2).sum() * dx, (np.abs(FTs) ** 2).sum() * FTs["freq_x"].spacing
```
roxyboy, line 661: I think by adding the phase and amplitude flags to power_spectrum, we could change this line to use that instead?

xrft/xrft.py
```diff
@@ -495,27 +495,27 @@ def idft(
         fft_fn = fftm.irfftn
```
roxyboy, line 495: I'm now wondering what would happen if `real` is not None but the phase and amplitude flags are set to True...

xrft/tests/test_xrft.py
lanougue: In test_parseval, line 584, 609, 623, 638, ... power spectrum is divided by the spacing(s). This normalisation should not be applied in case of true_phase/amplitude flags used.
Adding each time an "if" condition in case of flags being used starts to complicate the code. Don't you think we could have a better way ?

roxyboy:
> Adding true_phase and true_amplitude flags in power_spectrum and cross_spectrum is feasible (fft call should be transformed into dft call). However, output phase and amplitude will of course be modified and we should evaluate the impact of these modifications in these functions (which is probably also desirable too).

Considering that we already tests checking for Parseval when the flags were False (the current dft), I think the tests to be added should be only for when the flags are True. This would reduce the complexity of if statements.

xrft/tests/test_xrft.py
roxyboy: Considering that we already had tests checking for Parseval when the flags were False (the current dft), I think the tests to be added should be only for when the flags are True. This would reduce the complexity of if statements.

xrft/xrft.py
lanougue: I checked and it was not fine. This new PR should be ok now. I had to create an irfftfreq function which, strangely, is not a standard numpy function.

roxyboy: Thanks for looking into this. Can we also add a test for real dft with the flags being set to True?

lanougue:
I am simplifying and adding true_flags possibilities to (co, cross and phase of) spectrum.

I have remaing questions:
* Why cross_spectrum is returning only the real part ? I think that the imaginary part is as important as the real one. Shouln't we return the complex spectrum ?
* In computing cross_phase, I used xr.ufuncs.angle() to compute the phase. I think it automattically handle all dask and chunk stuff.
* What the return of power_spectrum with density flags set to False means ? Shouldn't we use the same syntax of "scipy.signal.periodogram" with a scaling flag set to "density" or "spectrum" ?

(@roxyboy : sorry for the failing checks, my local pytest is not working for now)

xrft/xrft.py
```diff
@@ -596,133 +605,19 @@ def power_spectrum(da, density=True, **kwargs):
...
     updated_dims = [
-        d for d in daft.dims if (d not in da.dims and "spacing" in daft[d].attrs)
-    ]  # checking spacing to be sure that it is not a segmented dimension
+        d for d in daft.dims if (d not in da.dims and "segment" not in d)
+    ]  # Transformed dimensions
     ps = np.abs(daft) ** 2
-    d = np.prod([float(ps.sizes[d] * ps[d].spacing) for d in updated_dims])
     if density:
-        ps /= d
+        fs = np.prod([float(ps[d].spacing) for d in updated_dims])
```
roxyboy, line 612: I think we should have an `if` statement here about the phase and amplitude flags. If density is set to True, it should satisfy Parseval regardless of the phase and amplitude flags.

xrft/xrft.py
```diff
@@ -737,96 +632,32 @@ def cross_spectrum(
...
+    ]  # Transformed dimensions
     cs = (daft1 * np.conj(daft2)).real
-
     if density:
-        cs /= (np.asarray(N).prod()) ** 2
-        for i in dim:
-            # cs /= daft1["freq_" + i + "_spacing"]
-            cs /= daft1["freq_" + i].spacing
-
+        fs = np.prod([float(ps[d].spacing) for d in updated_dims])
```
roxyboy, line 652: Same as in power_spectrum regarding the phase and amplitude flags.

xrft/xrft.py
```diff
@@ -879,32 +689,13 @@ def cross_phase(
...
 
-    if daft1.chunks and daft2.chunks:
-        _cross_phase = lambda a, b: dsar.angle(a * dsar.conj(b))
-    else:
-        _cross_phase = lambda a, b: np.angle(a * np.conj(b))
-    cp = xr.apply_ufunc(_cross_phase, daft1, daft2, dask="allowed")
+    if daft1.dims != daft2.dims:
+        raise ValueError("The two datasets have different dimensions")
+
+    cp = xr.ufuncs.angle(daft1 * np.conj(daft2))
```
roxyboy, line 698: I wasn't able to find the documentation about this on Google but could you provide the link for it so I can see what it's intended to do?

roxyboy (changes requested):

xrft/xrft.py
lanougue: Your are right, when density is True, true_flags do not matter. However, what density =False means ? Should we replace the density flag by a scaling flag as noted in my previous comment ?

roxyboy:
> * Why cross_spectrum is returning only the real part ? I think that the imaginary part is as important as the real one. Shouln't we return the complex spectrum ?
> * What the return of power_spectrum with density flags set to False means ? Shouldn't we use the same syntax of "scipy.signal.periodogram" with a scaling flag set to "density" or "spectrum" ?

Thanks for all the work @lanougue ! I think we're almost there. Regarding your point about cross spectrum, when I coded it up, I was only interested in the amplitude, which is why I ended up returning the real part only. If you think having the imaginary part would benefit users, I'm fine with returning the complex results.

As for the scaling flag, you're point makes sense but again, changing this flag would break the default users for current users... Essentially, `density=True` should be equivalent to `scaling=density`.

lanougue:
Yes, this is why I suggest to change this flag into a scaling flag (with a deprecated warning) since I beleive no one was using the density flag set to False

xrft/xrft.py
lanougue: http://xarray.pydata.org/en/stable/generated/xarray.ufuncs.angle.html

xrft/xrft.py
lanougue: I can not figure out a physical meaning of power_spectrum when density is set to False

roxyboy: It just means that it isn't normalized by the spacing, i.e. `scaling='spectrum'` if you want to use scipy's convention.

roxyboy: In order to not break the behavior, I vote for not changing this flag in this release. If we decide to change the flag to `scaling`, we would need to make both cases compatible, namely `scaling` and `density` flags and put a deprecation warning when `density` is called. `density=True` would be implemented as `scaling=density` and `density=False` would be implemented as `scaling=spectrum`.

roxyboy:
> I beleive that almost no one was using the density flag set to False. Can we just had a warning about deprecated "density" flag when the user sets it to False? Setting scaling="density' as default and removing density flag will ensure default behaviour being the same across version.
> Moreover, having the possibility to set scaling = "spectrum" seems to be used since it is in scipy.signal.periodogram function.

We can do this but we would still need to keep the `density` flag fully functional for this release so as not to break the behavior. If you want to change the flag to `scaling`, we would need to add a deprecation warning for when `density=True` or `density=False` is called.

lanougue:
@roxyboy, I have one failing tests that I can not resolve and that I think I should not resolve. In test_cross_phase_2d() line 516->544:
da1 and da2 have "y" coordinates that are all filled with "1" value. The spacing thus vanish for this dimension ! Fourier Transform over this dimension should thus not be possible.
However, dft do not complain because the spacing is still constant ! I think we should remove "y" coordinates or at least change it.
Can I change this test (while keeping its purpose of course) ?

roxyboy:
> @roxyboy, I have one failing tests that I can not resolve and that I think I should not resolve. In test_cross_phase_2d() line 516->544:
> da1 and da2 have "y" coordinates that are all filled with "1" value. The spacing thus vanish for this dimension ! Fourier Transform over this dimension should thus not be possible.
> However, dft do not complain because the spacing is still constant ! I think we should remove "y" coordinates or at least change it.
> Can I change this test (while keeping its purpose of course) ?

Good call. I agree, just change it to `range(Ny)` or something like that.

lanougue:
I also remove the Raise ValueError for the cross_phase in multiple dimensions. I do not see reason why it should not exist

xrft/tests/test_xrft.py
```diff
@@ -533,8 +533,6 @@ def test_cross_phase_2d(self, dask):
             dims=["y", "x"],
             coords={"y": np.arange(Ny), "x": x},
         )
-        with pytest.raises(ValueError):
-            xrft.cross_phase(da1, da2, dim=["y", "x"])
```
roxyboy, line 537: We should add a test that checks for two-dimensional cross_phase as we do for 1D below.

roxyboy (changes requested):

xrft/xrft.py
```diff
@@ -737,96 +632,32 @@ def cross_spectrum(
...
-def _cross_spectrum(daft1, daft2, dim, N, density):
+    updated_dims = [
+        d for d in daft1.dims if (d not in da1.dims and "segment" not in d)
+    ]  # Transformed dimensions
     cs = (daft1 * np.conj(daft2)).real
-
     if density:
-        cs /= (np.asarray(N).prod()) ** 2
-        for i in dim:
-            # cs /= daft1["freq_" + i + "_spacing"]
```
roxyboy, line 533: Can you remove this commented out line?

lanougue: It is noted in red color in github report and with "-" sign on left side. It is already removed. Right ?

roxyboy: Oh right, sorry got confused...

xrft/xrft.py
lanougue: density = True corresponds to scaling='density' BUT density=False does NOT corresponds to scaling='spectrum'.
I think that in the actual behaviour density = False do not correspond to any physical quantity. However, scaling='spectrum" does.
This is why I suggest to simply abandon density=False and add scaling="spectrum" possibility.
It also match with scipy periodogram and other spectra calculation.

I agree that it will break density=False behaviour but isn't it useless ? A deprecation warninng and maybe a suggestion to evaluate it from the output of density=True output could be a solution ?

roxyboy: Sorry, but breaking the behavior is a no go. If the `density=False` flag is bothering you, I'd recommend keeping the flag fully functional but with a deprecation flag and add the `scaling` flag in addition.

lanougue:
@roxyboy , I have modified test_cross_phase_2d to match its name.
I added one more test: test_cross_phase_true_phase_2d . For two signals with same values but with lagged coordinates, the offset should be recovered with the true_phase set to True in cross_phase.
I also think this behaviour will have to be documented

lanougue:
@roxyboy : To follow your recommandations, I added the scaling parameter in power_spectrum and cross_spectrum. Default scaling='density' will thus corresponds to actual default density=True. Density='False' will also be maintened (with warning).
Full consistency will thus be kept across versions.
If density flag is defined, a warning will be raised in both cases saying that density flag will be deprecated.
I also added scaling='spectrum' possibility

xrft/xrft.py
```diff
@@ -584,7 +584,65 @@ def idft(
...
+    ----------
+    da : `xarray.DataArray`
+        The data to be transformed
+    scaling : str, optional
+        If 'density', it will normalize the output to power spectral density
+        If 'spectrum', it will normalize the output to power spectrum
+    kwargs : dict : see xrft.dft for argument list
+    """
+
+    if "density" in kwargs:
```
roxyboy, line 606: Cool, thanks for adding this. Also, just another small typo: "not be maintained" in the warning msg.

xrft/xrft.py
```diff
@@ -584,7 +584,65 @@ def idft(
...
+    ]  # Transformed dimensions
+    ps = np.abs(daft) ** 2
+
+    if scaling == "density":
+        fs = np.prod([float(ps[d].spacing) for d in updated_dims])
+        ps *= fs
+    elif scaling == "spectrum":
+        fs = np.prod([float(ps[d].spacing) for d in updated_dims])
+        ps *= fs ** 2
+    else:  # Corresponds to density=False
```
roxyboy, line 636: I'm not sure if this else statement works. I think `scaling` needs to be set as None as default for this to work.

roxyboy (changes requested):

xrft/xrft.py
roxyboy: Or, `elif density == False` could be the argument.

lanougue: ok, I replaced it with a more robust version

lanougue:
If you think it is a good idea, we can continue to maintain the density=False capability by adding a possible value to scaling flag such as "false_density" (as done actually in this new commit)

roxyboy:
> If you think it is a good idea, we can continue to maintain the density=False capability by adding a possible value to scaling flag such as "false_density" (as done actually in this new commit)

It's not a matter of it being a good idea or not but rather a requirement to maintain all functionality with a deprecation cycle when maintaining a package. I get your point of `density=False` being useless (for us) but I have no idea who is using this `xrft` package and if a function is dropped without warning, some users may find that very troubling. As such, we need to maintain the `density=False` flag for this release. We can remove it in the next cycle.

lanougue:
Of course. My question was more on the maintainance on a long term purpose of density=False available result.
The question is : do we say it will disapear at some point in the future or do we say that if will be transformed in scaling='false_density' and be maintained.

xrft/xrft.py
```diff
@@ -584,7 +584,65 @@ def idft(
...
+        density = kwargs.pop("density")
+        msg = (
+            "density flag will be removed in future version of xrft.power_spectrum and replaced by scaling flag. "
+            + 'density=True should be replaced by scaling="density" and '
+            + "density=False will not be maintened"
+        )
+        warnings.warn(msg, FutureWarning)
+        if density:
+            if scaling != "density":
+                raise ValueError("Both scaling and density can not be defined")
```
roxyboy, line 616: An error message of "When density is set to True, scaling must be set to density." might be more helpful for the user. Also, we'd need to add a test for this ValueError.

roxyboy (changes requested):

xrft/xrft.py
roxyboy: And also a test for `scaling=spectrum`

roxyboy:
> Of course. My question was more on the maintainance on a long term purpose of density=False available result.
> The question is : do we say it will disapear at some point in the future or do we say that if will be transformed in scaling='false_density' and be maintained.

I think we can say that it'll be deprecated (meaning it'll disappear) in future releases.

xrft/xrft.py
lanougue: Ok for the test. However, the idea of the VelueError is to warn the user that he can not use both density flag and scaling flag. Your suggestion look like the contrary

lanougue: Maybe something like: "scaling flag and density flag can not be set simultenaously" ?

lanougue:
I add the test and change the location of the raise ValueError in order to be raised whatever density flag is

xrft/xrft.py
roxyboy: We should test a case where dft gives a ValueError when we would expect a ValueError. There are a few examples of this in test_xrft.py already so I'll try to refer the lines.

roxyboy: And I think `density=True` with `scaling='density'` should work for users. The cases where a ValueError should be raised are `density=False` with `scaling='density'` or `density=True` with `scaling='spectrum'`

xrft/xrft.py
```diff
@@ -606,14 +606,16 @@ def power_spectrum(da, scaling="density", **kwargs):
...
         msg = (
-            "density flag will be removed in future version of xrft.power_spectrum and replaced by scaling flag. "
+            "density flag will be deprecated in future version of xrft.power_spectrum and replaced by scaling flag. "
             + 'density=True should be replaced by scaling="density" and '
             + "density=False will not be maintained"
         )
         warnings.warn(msg, FutureWarning)
+        if scaling != "density":
+            raise ValueError(
+                "scaling flag and density flag can not be set simultenaously"
```
roxyboy, line 616: I think we should accept the case where they are both set at `density=True` with `scaling='spectrum'`

roxyboy (changes requested):

xrft/xrft.py
roxyboy: As in [here](https://github.com/xgcm/xrft/blob/a42f6b261abc8ab49d0baae9c2afb0d34fdcd895/xrft/tests/test_xrft.py#L96), we need to check that the function actually raises a ValueError as expected.

xrft/xrft.py
lanougue: What is it suppose to mean ? density=True matches with scaling='density', not scaling='spectrum'

lanougue: In fact, the code as it is now, accepts setting both density =True and scaling="density". Which is what you want, right ?

xrft/xrft.py
lanougue: As the code is now:
`density = True` with `scaling='density'` works.
`density = True` with `scaling='spectrum'` raise a ValueError
`density = False` with `scaling='spectrum'` raise a ValueError
`density = False` with `scaling='density'` only raise a Warning saying that scaling flag is ignored. I made it like this because scaling is set to "density" as default. Thus, setting `density = False` would always raise a ValueError even if scaling is not defined by the user.

I think that keeping it like this is acceptable and maintain a default power_spectrum behaviour between version. What do you think ?

A test is already added between line 571 and 575 in test_xrft

xrft/xrft.py
roxyboy: > In fact, the code as it is now, accepts setting both density =True and scaling="density". Which is what you want, right ?

Yes :) But the Value Error could be interpreted by the user as this won't work as `density=True` and `scaling='density'` is also setting the flags simultaneously.

xrft/tests/test_xrft.py
```diff
@@ -568,6 +568,12 @@ def test_cross_phase_true_phase_2d(self, dask):
         )  # Offset in [-pi, pi]
         npt.assert_almost_equal(offset, phase_offset)
 
+    def test_scaling_density_flag(self):
+        N = 25
+        s = xr.DataArray(np.random.rand(N), dims="x", coords={"x": range(N)})
+        with pytest.raises(ValueError):
+            ps = xrft.power_spectrum(s, dim="x", scaling="spectrum", density=True)
```
roxyboy, line 575: Great, thanks for adding this test. Could you also add a case for `density=False` and `scaling='density'`?

roxyboy (changes requested):

xrft/tests/test_xrft.py
lanougue: In this condition, only a warning will be raised. Should we test for a warning ?

roxyboy: Ah ok, no need for testing the warning then.

lanougue: Ok, it think it should raise an error too. I will correct that

roxyboy: Can you add the case for `density=False` and `scaling='spectrum'`?

roxyboy: Actually, since the default is not going to be set as `scaling='density'`, I think the case of `densitt=False` and `scaling='density'` should only be a warning and not an error.

roxyboy: I like your original idea of when `density=False`, the `scaling` flag being ignored.

lanougue: Ok, so let's go for that. I just add the other test for the raised error

xrft/tests/test_xrft.py
```diff
@@ -568,11 +568,12 @@ def test_cross_phase_true_phase_2d(self, dask):
...
         npt.assert_almost_equal(offset, phase_offset)
 
-    def test_scaling_density_flag(self):
+    @pytest.mark.parametrize("density", [False, True])
+    def test_scaling_density_flag(self, density):
         N = 25
         s = xr.DataArray(np.random.rand(N), dims="x", coords={"x": range(N)})
         with pytest.raises(ValueError):
-            ps = xrft.power_spectrum(s, dim="x", scaling="spectrum", density=True)
+            ps = xrft.power_spectrum(s, dim="x", scaling="spectrum", density=density)
```
roxyboy, line 576: This tests for raising a ValueError when `density=False` but... it should only raise a warning because `scaling` will be ignored right?

lanougue: The warning is raised only when `scaling = "density"`, not "spectrum". The test for ValueError is realized only with `scaling = "spectrum"`

lanougue: When `scaling="density"`, either density is et to True and the code works fine since everything is consistent, either density is set to False and the warning is raised

roxyboy: Gotcha.

xrft/xrft.py
```diff
@@ -365,82 +419,234 @@ def power_spectrum(
...
+            ):  # means that input is on fftshifted grid
+                daft = daft.sortby(d)  # reordering the input
+            else:
+                raise ValueError(
+                    "Can't take Fourier transform because "
+                    "coodinate %s is not evenly spaced" % d
+                )
+        if np.abs(l) > spacing_tol:
+            raise ValueError(
+                "Inverse Fourier Transform can not be computed because coordinate %s is not centered on zero frequency"
```
roxyboy, line 534: Sorry for nit picking. Do we have a test to check for this Value Error?

lanougue: Nope, I can make one.

I think that we aslo should have a ValueError when coordinates has values that all are the same. The returned error: division by zero (the spacing) is not really explicit.

lanougue: Both done

lanougue:
Even if there is no need of compatibility of idft with current xrft version, I just restore idft default behaviour with true_flags to False to be consistent with dft. (I set them to true in the commit before). I changed the message of the warning to be more explicit

xrft/xrft.py
```diff
@@ -455,6 +457,10 @@ def idft(
         The output of the Inverse Fourier transformation, with appropriate dimensions.
     """
 
+    if not true_phase and not true_amplitude:
+        msg = "Flags true_phase and true_amplitude will be set to True in future versions of xrft.idft to preserve the theoretical phasing and amplitude of IFT. Consider using xrft.ifft to ensure future compatibility with numpy.ifft like behavior and to deactivate this warning."
```
roxyboy, line 461: Maybe we can spell out IFT as inverse FT to be user friendly?

xrft/xrft.py
roxyboy: Is this resolved? I just want to make sure so that the msg doesn't confuse the user.

roxyboy:
I think this is ready to merge. In hindsight, we probably should've split up this massive PR into multiple short PRs but "c'est la vie". @rabernat Can you give this PR a final approval?

roxyboy:
Thanks for you patience and all the work @lanougue !

lanougue:
Thank you too for reviewing all these modifications ! 

A last thing. In cross_phase, I set true_phase=False as default to keep the default behaviour across version
 In beleive that in future we should probably change it to true because it takes into account the coordinates to evaluate the correct cross_phase.
I think we should had a warning here too.

roxyboy:
> A last thing. In cross_phase, I set true_phase=False as default to keep the default behaviour across version
> In beleive that in future we should probably change it to true because it takes into account the coordinates to evaluate the correct cross_phase.
> I think we should had a warning here too.

I agree, a warning would be nice.

lanougue:
I added warnings for both cross_spectrum and cross_phase. All true_flags default behaviour is now only supported by dft and idft functions ensuring compatibility across all functions which depend on them.
Code looks more clear to me.

xrft/xrft.py
```diff
@@ -742,7 +749,14 @@ def cross_phase(da1, da2, true_phase=False, **kwargs):
         The data to be transformed
     kwargs : dict : see xrft.dft for argument list
     """
-    cp = xr.ufuncs.angle(cross_spectrum(da1, da2, true_phase=true_phase, **kwargs))
+    if "true_phase" not in kwargs:
+        msg = (
+            "true_phase flag will be set to True in future version of xrft.dft possibly impacting cross_phase output. "
+            + "Set explicitely true_phase = False in cross_phase arguments list to ensure future compatibility."
```
roxyboy, line 755: I'd add "future compatibility for numpy-like behavior where the coordinates are disregarded." As it stands, it sounds like setting the flag to False is better for future compatibility in general.

xrft/xrft.py
```diff
@@ -676,6 +676,13 @@ def cross_spectrum(da1, da2, scaling="density", true_phase=False, **kwargs):
     kwargs : dict : see xrft.dft for argument list
     """
 
+    if "true_phase" not in kwargs:
+        msg = (
+            "true_phase flag will be set to True in future version of xrft.dft possibly impacting cross_spectrum output. "
+            + "Set explicitely true_phase = False in cross_spectrum arguments list to ensure future compatibility."
```
roxyboy, line 682: Same point as below regarding "future compatibility".

roxyboy (changes requested):

xrft/xrft.py
lanougue: I prefer "disregarded" to "discarded"

xrft/xrft.py
roxyboy: Also a small typo for "simultaneously."

xrft/xrft.py
roxyboy: Sure, whichever works for me.

xrft/xrft.py
lanougue: Much better indeed !

xrft/xrft.py
roxyboy: How about "scaling cannot be set to 'spectrum' when density is defined."?

lanougue:
Loooong PR ! :) Happy that it reaches the end

xrft/xrft.py
roxyboy: @lanougue Last fix please...

lanougue: Actually, the code only accepts to set both scaling='density' and density='True' because they mean the same thing.
Other configuration with both defined should raise a ValueError.
For now, we only makes the exception of density=False and scaling="desnity" which only raise a warning saying that scaling is ignored

lanougue: I think the user should never be able to set both scaling and density since it leads to severe misunderstanding on what the code should return.

roxyboy: Ok, but my initial claim still holds. The Error message as it stands states that it won't accept the case of scaling='density' and density=True either.

roxyboy: In essence, we're only returning a ValueError when `density` is called no matter it be True or False when `scaling='spectrum'`

roxyboy: So, we should make the error message match this.

lanougue: Yes. As it, even if the error message claims that, it still  work because 1) We allow it 2) they are equivalent.
What do you suggest ?

roxyboy: I recommend the error to be: "scaling cannot be set to 'spectrum' when density is called." as this is what what's actually happening.

roxyboy: *when the density flag is called.

lanougue: I suggest that, for clarity, an error should always be raised if both flags are defined, no matter their value.

roxyboy: If we do this, we'd need to set the default of scaling to None as setting scaling='density' as the default would always raise an error when the density flag is called and this would break the behavior. So I do not agree with this.

lanougue: You are right. But I am sure I can find a workaround to have the default scaling to density without raising the error. I just need to check if the user set it explicitely or not in the argument list

roxyboy: What you're proposing sounds like a lot of extra `if` statements for a feature that'll be deprecated in future cycles. I'm not sure why you're so against just clarifying the error message of what it's doing...

roxyboy: We already have in the warning message above a more detailed explanation whenever the density flag is called.

lanougue: I do not really like it because it sounds that in some cases we are allowed to set both flags and sometime we are not allowed, depending on their values. Which is strange for me.
I prefer to make the user choose one of the two flags

lanougue: Having both flags set simultaneously would mean that the user do not know how to choose them

lanougue: Do I give a try ? It will gain in clarity with simple error message I think

roxyboy: Ok, you seem willing to try it so, why not.

xrft/xrft.py
```diff
@@ -612,24 +612,24 @@ def power_spectrum(da, scaling="density", **kwargs):
...
-    if "density" in kwargs:
-        density = kwargs.pop("density")
-        msg = (
-            "density flag will be deprecated in future version of xrft.power_spectrum and replaced by scaling flag. "
-            + 'density=True should be replaced by scaling="density" and '
-            + "density=False will not be maintained"
-        )
-        warnings.warn(msg, FutureWarning)
-        if scaling != "density":
+    if "scaling" in kwargs:
```
roxyboy, line 615: The problem with hiding `scaling` in kwargs in my opinion is that the user will have a hard time knowing that the flag even exists as it's not in xrft.dft...

xrft/xrft.py
lanougue: Sorry for being a bit stubborn, but I think it is more clear for a user point of view.
code is even shorter ! :)

xrft/xrft.py
roxyboy: If we're going to relegate `scaling` to kwargs, we need to add more explanation to the docstring of the function to explain what it takes in addition to kwargs accepted in xrft.dft

lanougue: Yes, I agree. It is in the description of the argument list, but I do not know another way.
Do we have another way to check if the user set an argument explicitely or if it is the default value ?

lanougue: Maybe some hope with that:
https://stackoverflow.com/questions/14749328/how-to-check-whether-optional-function-parameter-is-set

xrft/xrft.py
```diff
@@ -612,24 +612,24 @@ def power_spectrum(da, scaling="density", **kwargs):
     kwargs : dict : see xrft.dft for argument list
```
roxyboy, line 612: I think we should list the possible keyword arguments that are not in xrft.dft as they do [here](https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs)

roxyboy (changes requested):

xrft/xrft.py
roxyboy: So the list should have `scaling` and `density`.

lanougue: Ok, Following you link recommandations, `dim` should also appear because it is always used

roxyboy: Agreed.

lanougue: I agree that function signature is important. I 'll try to find a way

lanougue: What do you think about this solution:

We keep `scaling=True` in the signature of the function. It define its default behaviour (which continue the same as the actual version).
'density' flag stays in kwargs for two reasons:
1) it will be deprecated
2) This will not encourage new user to use it

If density is still manually set by the user (easily detectable since it is in the kwargs), we raise a warning saying that scaling flag is simply ignored.
This warning thus occurs when density if defined by the user and whatever density and scaling values are. This ensure compatibility and smooth transition.

I think it fullfil all our needs

xrft/xrft.py
```diff
@@ -612,25 +612,21 @@ def power_spectrum(da, **kwargs):
...
-            scaling = "density"
+    if "density" in kwargs:
+        density = kwargs.pop("density")
+        msg = (
+            "density flag will be deprecated in future version of xrft.power_spectrum and replaced by scaling flag. "
+            + 'density=True should be replaced by scaling="density" and '
+            + "density=False will not be maintained"
+        )
+        warnings.warn(msg, FutureWarning)
+        msg = "scaling flag is ignored !"
```
roxyboy, line 623: Can we just combine this warning with the one above?

xrft/tests/test_xrft.py
```diff
@@ -568,14 +568,14 @@ def test_cross_phase_true_phase_2d(self, dask):
...
 
-    @pytest.mark.parametrize("density", [False, True])
-    def test_scaling_density_flag(self, density):
-        N = 25
-        s = xr.DataArray(np.random.rand(N), dims="x", coords={"x": range(N)})
-        with pytest.raises(ValueError):
-            ps = xrft.power_spectrum(s, dim="x", scaling="spectrum", density=density)
-        with pytest.raises(ValueError):
-            ps = xrft.power_spectrum(s, dim="x", scaling="density", density=density)
+    # @pytest.mark.parametrize("density", [False, True])
```
roxyboy, line 571: Can you remove the commented out lines?

lanougue:
Happy new year 2021. May xrft have a long life !
@roxyboy , @rabernat . What about this PR ? followed by a new release ?

xrft/xrft.py
rabernat: Has this issue been resolved?

rabernat:
Sorry @lanougue we have neglected your PR for so long. From my point of view, everything looks good to merge.

@roxyboy do you agree?

xrft/xrft.py
rabernat: I support the idea of deprecating the `density=False` option and moving towards the `scaling` parameter. But we need to do this via a deprecation cycle. That is what appears to be implemented now.

xrft/xrft.py
lanougue: I think so. "scaling" now appears in argument list as expected. If the user keep using "density", it will work but will raise a deprecation warning

xrft/xrft.py
lanougue: Yes, this is what now appears in the last version of the PR

roxyboy:
@rabernat Yes, I agree with merging this PR :)

PR #132, opened 2021-01-29
numpy_groupies is now required

cspencerjones:
It seems as though `numpy_groupies` is now required in order to run `import xrft` (rather than an extra), so I suggest moving it to `INSTALL_REQUIRES`. 

(suggestions for other possibilities are welcome)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,8 +21,8 @@ CLASSIFIERS = [
     "Topic :: Scientific/Engineering",
 ]
 
-INSTALL_REQUIRES = ["xarray", "dask", "numpy", "pandas", "scipy"]
-EXTRAS_REQUIRE = ["cftime", "numpy_groupies"]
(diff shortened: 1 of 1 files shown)


rabernat:
Interesting. Yes, I guess @roxyboy added this dependency in d6cbc0ab383ab2c6406b2145601eb82c016b448f but never added the dependency.

roxyboy:
Yea, I've been thinking about moving the call `import numpy_groupies` inside the `isotropic` functions as it's only there where it's used

roxyboy:
I think I'd like to keep it as extra requirement as `numpy_groupies` has quite picky dependencies.
I'll work on a PR.

roxyboy:
PR #133 should fix this ;)

cspencerjones:
OK, no worries, I'll close this.

PR #133, opened 2021-01-29
Making numpy_groupies an extra requirement

roxyboy:
This addresses PR #132 . I moved `import numpy_groupies` to only when `isotropic` functions are called.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,8 +21,8 @@ CLASSIFIERS = [
     "Topic :: Scientific/Engineering",
 ]
 
-INSTALL_REQUIRES = ["xarray", "dask", "numpy", "pandas", "scipy"]
-EXTRAS_REQUIRE = ["cftime", "numpy_groupies"]
+INSTALL_REQUIRES = ["xarray", "dask", "numpy", "pandas", "scipy", "numpy_groupies"]
+EXTRAS_REQUIRE = ["cftime"]
 SETUP_REQUIRES = ["pytest-runner"]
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -21,8 +21,8 @@
     "Topic :: Scientific/Engineering",
 ]
 
-INSTALL_REQUIRES = ["xarray", "dask", "numpy", "pandas", "scipy"]
-EXTRAS_REQUIRE = ["cftime", "numpy_groupies"]
+INSTALL_REQUIRES = ["xarray", "dask", "numpy", "pandas", "scipy", "numpy_groupies"]
```
rabernat, line 24: By placing numpy_groupies here, it is now a required dependency...

xrft/xrft.py
```diff
@@ -782,6 +781,9 @@ def _binned_agg(
     dtype,
 ) -> np.ndarray:
     """NumPy helper function for aggregating over bins."""
+
+    import numpy_groupies
```
rabernat, line 785: ...so there is no point in only importing it here. You might as well leave it at the top, because it will always be installed.

rabernat (commented):
The two changes in the PR seem contradictory.

setup.py
roxyboy: Good point, I'll put it back under extra_require

roxyboy: I think I messed up my git so I'm gonna make a new PR, sorry...

roxyboy:
I think I messed up my git so I'm gonna make a new PR, sorry...

PR #134, opened 2021-01-29
Making numpy_groupies an extra requirement

roxyboy:
I moved the call `import numpy_groupies` only to when `isotropic` functions are called.

### DIFF
--- a/xrft/xrft.py
+++ b/xrft/xrft.py
@@ -782,7 +782,12 @@ def _binned_agg(
 ) -> np.ndarray:
     """NumPy helper function for aggregating over bins."""
 
-    import numpy_groupies
+    try:
+        import numpy_groupies
+    except ImportError:
+        raise ImportError(
+            "This function requires the `numpy_groupies` package to be installed. Please install it with pip or conda."
+        )
 
     mask = np.logical_not(np.isnan(indices))
     int_indices = indices[mask].astype(int)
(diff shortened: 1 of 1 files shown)


xrft/xrft.py
```diff
@@ -782,6 +781,9 @@ def _binned_agg(
     dtype,
 ) -> np.ndarray:
     """NumPy helper function for aggregating over bins."""
+
+    import numpy_groupies
```
rabernat, line 785:
```suggestion
    try:
        import numpy_groupies
    except ImportError:
        raise ImportError("This function requires the `numpy_groupies` package to be installed. Please install it with pip or conda.")
```

rabernat (commented):
It would be nice to have a more verbose error message.

xrft/xrft.py
roxyboy: I added the import error. I've never made a test to catch such error but do we need one?

rabernat: I think it helps the user understand better. Just imagine you are a python n00b and you try this function and you get a random import error for a package you have never heard of!

roxyboy: I agree with your point. I'm just wondering if I'd need to add another test in `test_xrft.py` to check for this `ImportError`...

rabernat: Not necessary IMO.

roxyboy: Ok cool. So we can merge this?

Issue #105, opened 2020-08-26
Bug with spectra when using real fft


dougiesquire:
I think there is a bug with the power spectrum estimates when a `real` dimension is provided. 

For real fft, Hermitian symmetry is exploited to compute only the non-negative frequency terms. But we can't ignore the energy in the negative frequency terms when computing spectra. To account for this, we need to multiply by 2 the energy of all but the first (DC) mode (and last mode if the signal length even - see https://numpy.org/doc/stable/reference/generated/numpy.fft.rfft.html).

I can demonstrate the issue and the fix through comparison to `scipy.signal.periodogram`:
```python
import xrft
import numpy as np
import scipy.signal
import xarray as xr
import matplotlib.pyplot as plt 

x = xr.DataArray(np.random.normal(0, 1, size=100), 
                 dims=['t'], 
                 coords=[range(0,100)])


f_scipy, p_scipy = scipy.signal.periodogram(x.values,
                                            window='rectangular')
p_xrft = xrft.xrft.power_spectrum(x, 
                                  dim=['t'], 
                                  real='t',
                                  detrend='constant')

# For even length signals the last mode includes both the +ve and -ve contributions
correction = np.ones_like(f_scipy)
correction[1:-1] = 2

plt.plot(f_scipy, p_scipy,
         label='scipy psd')
plt.plot(p_xrft.freq_t, p_xrft, 
         color='C2', label='xrft psd')
plt.plot(p_xrft.freq_t, correction*p_xrft, 
         linestyle='--', color='C1', label='corrected xrft psd')
plt.xlabel('freq')
plt.ylabel('psd')
plt.legend()
```
![delete-3](https://user-images.githubusercontent.com/42455466/91303429-5d7c9d00-e7eb-11ea-982b-0d5a598cc0e2.png)

roxyboy:
These are all really great fixes! Thanks for putting in the time.

lanougue:
Hi @dougiesquire, @roxyboy 

I think this is not an xrft nor scipy periodogram bug. In scipy.signal.periodogram(), there is a default parameter "return_onesided" which is set to "True" as default. xrft power_spectrum() returns a two_sided spectrum.
If you set return_onesided parameter to False in scipy.signal.periodogram(), you should get a two_sided spectrum whose amplitude match xrft.power_spectrum.
Having return_onesided=True in scipy periodrogram() double the amplitude to compensate the "one sided".

lanougue:
[...]

PR #135, opened 2021-01-30
More examples for documentation

roxyboy:
I started a PR to have more examples for the documentation. I basically compiled all the examples raised by other contributors. This PR address issues #86 and #105 .

Issue #104, opened 2020-08-26
Correction for reduced energy when windowing


dougiesquire:
Applying window weighting (e.g. Hann) to a signal reduces its energy. I think it's common to want to correct for this. For example, it appears that such a correction is applied by default in `scipy.signal.welch`:

```python
import xrft
import numpy as np
import scipy.signal
import xarray as xr
import matplotlib.pyplot as plt 

x = xr.DataArray(np.random.normal(0, 1, size=10000), 
                 dims=['t'], 
                 coords=[range(0,10000)])
n_segment = 100

f_scipy, p_scipy = scipy.signal.welch(x.values,
                                      window='hann',
                                      nperseg=n_segment,
                                      noverlap=0,
                                      return_onesided=False)
                
p_xrft = xrft.xrft.power_spectrum(x.chunk({'t': n_segment}), 
                                  dim=['t'],
                                  detrend='constant',
                                  window=True,
                                  chunks_to_segments=True).mean('t_segment')

# Correction tends to 8/3 for infinitely long Hann window
p_xrft_corrected = (1 / np.mean(np.hanning(n_segment)**2))*p_xrft 

plt.plot(f_scipy, p_scipy, 
         label='scipy psd')
plt.plot(p_xrft.freq_t, p_xrft, 
         color='C2', label='xrft psd')
plt.plot(p_xrft.freq_t, p_xrft_corrected, 
         linestyle='--', color='C1', label='corrected xrft psd')
plt.ylim(0,1.6)
plt.xlabel('freq')
plt.ylabel('psd')
plt.legend()
```
![delete-2](https://user-images.githubusercontent.com/42455466/91297889-6ddc4a00-e7e2-11ea-95a8-28a3dfb8e2de.png)

Is there an appetite (beyond my own) to have an option to apply such a correction when using `window=True` in `xrft`?

navidcy:
Was this fixed or should it be still an open issue?

roxyboy:
No, do you want to start a PR @navidcy ?

navidcy:
I might but perhaps next week ;)

roxyboy:
@dougiesquire I haven't been able to satisfy the Parseval's equation when I apply the correction you've suggested, i.e. `p_xrft_corrected.mean('freq_t) = ((x_stacked - x_stacked.mean(['t']))**2).mean(['t_segment']).mean('t)` where `x_stacked` is `x` stacked into n_segment chunks. Have you been able to sort this out?

[...]

PR #136, opened 2021-02-06
Making FFT conventions similar to scipy

roxyboy:
A PR to double the power and cross spectrum when `real` is called in order to adhere with `scipy.signal.periodogram(return_onesided=True)` convention. This addresses issue #105 .

### DIFF
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -407,6 +407,11 @@ class TestSpectrum(object):
         npt.assert_almost_equal(ps.values, np.real(daft * np.conj(daft)))
         npt.assert_almost_equal(np.ma.masked_invalid(ps).mask.sum(), 0.0)
 
+        with pytest.raises(ValueError):
+            xrft.power_spectrum(
+                da, dim=["y", "x"], window=False, correct_amplitude=True
+            )
+
     @pytest.mark.parametrize("dask", [False, True])
     def test_cross_spectrum(self, dask):
         """Test the cross spectrum function"""
@@ -451,6 +456,9 @@ class TestSpectrum(object):
         test /= dk ** 2
         npt.assert_almost_equal(cs.values, test)
         npt.assert_almost_equal(np.ma.masked_invalid(cs).mask.sum(), 0.0)
+
+        with pytest.raises(ValueError):
+            xrft.cross_spectrum(da, da2, dim=dim, window=False, correct_amplitude=True)
 
     def test_spectrum_dim(self):
         N = 16
@@ -640,10 +648,14 @@ def test_parseval(chunks_to_segments):
         dims=dim,
         coords=da.coords,
     )
+
     # Check that the (rectangular) integral of the spectrum matches the windowed variance
+    # for i in dim:
+    #     da_corrected = (da_prime * window) ** 2 / np.mean(np.hanning(N / n_segments)**2)
     npt.assert_almost_equal(
         (1 / delta_xy) * ps.mean(fftdim).values,
         ((da_prime * window) ** 2).mean(dim).values,
+        # da_corrected.mean(dim).values,
         decimal=5,
     )
 
@@ -658,6 +670,7 @@ def test_parseval(chunks_to_segments):
     npt.assert_almost_equal(
         (1 / delta_xy) * cs.mean(fftdim).values,
         ((da_prime * window) * (da2_prime * window)).mean(dim).values,
+        # ((da_prime) * (da2_prime)).mean(dim).values,
         decimal=5,
     )
 
@@ -673,7 +686,9 @@ def test_parseval(chunks_to_segments):
         npt.assert_almost_equal(
             (1 / delta_xy) * ps[0].values.mean(),
             (
-                (xrft.detrend(d3d, dim, detrend_type="linear")[0].values * window) ** 2
+                (xrft.detrend(d3d, dim, detrend_type="linear")[0].values * window)
+                ** 2
+                # (xrft.detrend(d3d, dim, detrend_type="linear")[0].values) ** 2
             ).mean(),
             decimal=5,
         )
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -409,7 +409,7 @@ class TestSpectrum(object):
 
         with pytest.raises(ValueError):
             xrft.power_spectrum(
-                da, dim=["y", "x"], window=False, correct_amplitude=True
+                da, dim=["y", "x"], window=False, boost_amplitude=True
             )
 
     @pytest.mark.parametrize("dask", [False, True])
@@ -458,7 +458,7 @@ class TestSpectrum(object):
         npt.assert_almost_equal(np.ma.masked_invalid(cs).mask.sum(), 0.0)
 
         with pytest.raises(ValueError):
-            xrft.cross_spectrum(da, da2, dim=dim, window=False, correct_amplitude=True)
+            xrft.cross_spectrum(da, da2, dim=dim, window=False, boost_amplitude=True)
 
     def test_spectrum_dim(self):
         N = 16
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -387,7 +387,7 @@ class TestSpectrum(object):
                 dim=["t"],
                 window=window_type,
                 chunks_to_segments=True,
-                boost_windowed_amplitude=True,
+                window_correction=True,
             ).mean("t_segment")
 
             npt.assert_allclose(
@@ -442,9 +442,7 @@ class TestSpectrum(object):
         npt.assert_almost_equal(np.ma.masked_invalid(ps).mask.sum(), 0.0)
 
         with pytest.raises(ValueError):
-            xrft.power_spectrum(
-                da, dim=["y", "x"], window=None, boost_windowed_amplitude=True
-            )
+            xrft.power_spectrum(da, dim=["y", "x"], window=None, window_correction=True)
 
     @pytest.mark.parametrize("dask", [False, True])
     def test_cross_spectrum(self, dask):
@@ -492,9 +490,7 @@ class TestSpectrum(object):
         npt.assert_almost_equal(np.ma.masked_invalid(cs).mask.sum(), 0.0)
 
         with pytest.raises(ValueError):
-            xrft.cross_spectrum(
-                da, da2, dim=dim, window=None, boost_windowed_amplitude=True
-            )
+            xrft.cross_spectrum(da, da2, dim=dim, window=None, window_correction=True)
 
     def test_spectrum_dim(self):
         N = 16
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -650,14 +650,22 @@ def test_parseval(chunks_to_segments):
     )
 
     # Check that the (rectangular) integral of the spectrum matches the windowed variance
-    # for i in dim:
-    #     da_corrected = (da_prime * window) ** 2 / np.mean(np.hanning(N / n_segments)**2)
     npt.assert_almost_equal(
         (1 / delta_xy) * ps.mean(fftdim).values,
         ((da_prime * window) ** 2).mean(dim).values,
-        # da_corrected.mean(dim).values,
         decimal=5,
     )
+
+    # ps = xrft.power_spectrum(
+    #     da, window=True, detrend="constant", chunks_to_segments=chunks_to_segments, correct_amplitude=True
+    # )
+    # da_corrected = (da_prime * window) ** 2 / (window**2).mean()
+    # npt.assert_almost_equal(
+    #     (1 / delta_xy) * ps.mean(fftdim).values,
+    #     ((da_prime * window) ** 2).mean(dim).values,
+    #     da_corrected.mean(dim).values,
+    #     decimal=2,
+    # )
 
     ### Test Parseval's theorem for cross_spectrum with `window=True` and detrend='constant'
     cs = xrft.cross_spectrum(
@@ -670,7 +678,6 @@ def test_parseval(chunks_to_segments):
     npt.assert_almost_equal(
         (1 / delta_xy) * cs.mean(fftdim).values,
         ((da_prime * window) * (da2_prime * window)).mean(dim).values,
-        # ((da_prime) * (da2_prime)).mean(dim).values,
         decimal=5,
     )
 
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -367,7 +367,7 @@ class TestSpectrum(object):
         f_scipy, p_scipy = sps.periodogram(
             da.values, window="rectangular", return_onesided=True
         )
-        ps = xrft.xrft.power_spectrum(da, dim="x", real="x", detrend="constant")
+        ps = xrft.power_spectrum(da, dim="x", real="x", detrend="constant")
         npt.assert_almost_equal(ps.values, p_scipy)
 
         A = 20
@@ -382,7 +382,7 @@ class TestSpectrum(object):
             # see https://github.com/scipy/scipy/blob/master/scipy/signal/tests/test_spectral.py#L485
 
             x_da = xr.DataArray(x, coords=[tt], dims=["t"]).chunk({"t": n_segments})
-            ps = xrft.xrft.power_spectrum(
+            ps = xrft.power_spectrum(
                 x_da,
                 dim="t",
                 window=window_type,
@@ -396,7 +396,7 @@ class TestSpectrum(object):
                 rtol=1e-3,
             )
 
-            ps = xrft.xrft.power_spectrum(
+            ps = xrft.power_spectrum(
                 x_da,
                 dim="t",
                 window=window_type,
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -123,7 +123,10 @@ class TestFFTImag(object):
 
         ft = xrft.fft(da, shift=False, window="hann", detrend="constant")
         dim = da.dims
-        window = sps.windows.hann(N) * sps.windows.hann(N)[:, np.newaxis]
+        window = (
+            sps.windows.hann(N, sym=False)
+            * sps.windows.hann(N, sym=False)[:, np.newaxis]
+        )
         da_prime = (da - da.mean(dim=dim)).values
         npt.assert_almost_equal(ft.values, np.fft.fftn(da_prime * window))
 
@@ -367,33 +370,39 @@ class TestSpectrum(object):
         ps = xrft.xrft.power_spectrum(da, dim="x", real="x", detrend="constant")
         npt.assert_almost_equal(ps.values, p_scipy)
 
-        if dask:
-            N = 1000
-            da = xr.DataArray(
-                np.random.rand(N),
-                dims=["x"],
-                coords={
-                    "x": range(N),
-                },
-            )
-            n_segment = 100
-            f_scipy, p_scipy = sps.welch(
-                da.values,
-                window="hann",
-                nperseg=n_segment,
+        A = 20
+        fs = 1e4
+        n_segments = int(fs // 10)
+        fsig = 300
+        ii = int(fsig * n_segments // fs)  # Freq index of fsig
+
+        tt = np.arange(fs) / fs
+        x = A * np.sin(2 * np.pi * fsig * tt)
+        for window_type in ["hann", "bartlett", "flattop"]:
+            # see https://github.com/scipy/scipy/blob/master/scipy/signal/tests/test_spectral.py#L485
+            freq, p_scipy = sps.welch(
+                x,
+                fs=fs,
+                nperseg=n_segments,
+                window=window_type,
                 noverlap=0,
                 return_onesided=False,
+                scaling="density",
             )
+
+            x_da = xr.DataArray(x, coords=[tt], dims=["t"]).chunk({"t": n_segments})
             ps = xrft.xrft.power_spectrum(
-                da.chunk({"x": n_segment}),
-                dim="x",
-                detrend="constant",
-                window="hann",
-                chunks_to_segments=True,
-            ).mean("x_segment")
-            ps_corrected = ps / np.mean(sps.windows.hann(n_segment) ** 2)
+                x_da, dim=["t"], window=window_type, chunks_to_segments=True
+            ).mean("t_segment")
+
+            corr = 1 / np.mean(
+                getattr(sps.windows, window_type)(n_segments, sym=False) ** 2
+            )
+
             npt.assert_allclose(
-                ps_corrected.values, np.fft.fftshift(p_scipy), atol=1e-2
+                np.sqrt(np.trapz(corr * ps.values, ps.freq_t.values)),
+                A * np.sqrt(2) / 2,
+                rtol=1e-3,
             )
 
         da = xr.DataArray(
@@ -682,8 +691,8 @@ def test_parseval(chunks_to_segments):
     window = xr.DataArray(
         np.tile(
             # np.hanning(N / n_segments) * np.hanning(N / n_segments)[:, np.newaxis],
-            sps.windows.hann(int(N / n_segments))
-            * sps.windows.hann(int(N / n_segments))[:, np.newaxis],
+            sps.windows.hann(int(N / n_segments), sym=False)
+            * sps.windows.hann(int(N / n_segments), sym=False)[:, np.newaxis],
             (n_segments, n_segments),
         ),
         dims=dim,
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -123,7 +123,6 @@ class TestFFTImag(object):
 
         ft = xrft.fft(da, shift=False, window="hann", detrend="constant")
         dim = da.dims
-        # window = np.hanning(N) * np.hanning(N)[:, np.newaxis]
         window = sps.windows.hann(N) * sps.windows.hann(N)[:, np.newaxis]
         da_prime = (da - da.mean(dim=dim)).values
         npt.assert_almost_equal(ft.values, np.fft.fftn(da_prime * window))
@@ -693,17 +692,6 @@ def test_parseval(chunks_to_segments):
         ((da_prime * window) ** 2).mean(dim).values,
         decimal=5,
     )
-
-    # ps = xrft.power_spectrum(
-    #     da, window=True, detrend="constant", chunks_to_segments=chunks_to_segments, correct_amplitude=True
-    # )
-    # da_corrected = (da_prime * window) ** 2 / (window**2).mean()
-    # npt.assert_almost_equal(
-    #     (1 / delta_xy) * ps.mean(fftdim).values,
-    #     ((da_prime * window) ** 2).mean(dim).values,
-    #     da_corrected.mean(dim).values,
-    #     decimal=2,
-    # )
 
     ### Test Parseval's theorem for cross_spectrum with `window=True` and detrend='constant'
     cs = xrft.cross_spectrum(
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -383,15 +383,15 @@ class TestSpectrum(object):
 
             x_da = xr.DataArray(x, coords=[tt], dims=["t"]).chunk({"t": n_segments})
             ps = xrft.xrft.power_spectrum(
-                x_da, dim=["t"], window=window_type, chunks_to_segments=True
+                x_da,
+                dim=["t"],
+                window=window_type,
+                chunks_to_segments=True,
+                boost_windowed_amplitude=True,
             ).mean("t_segment")
 
-            corr = 1 / np.mean(
-                getattr(sps.windows, window_type)(n_segments, sym=False) ** 2
-            )
-
             npt.assert_allclose(
-                np.sqrt(np.trapz(corr * ps.values, ps.freq_t.values)),
+                np.sqrt(np.trapz(ps.values, ps.freq_t.values)),
                 A * np.sqrt(2) / 2,
                 rtol=1e-3,
             )
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -212,7 +212,7 @@ class TestfftReal(object):
         dx = float(da.x[1] - da.x[0]) if "x" in da.dims else 1
 
         # defaults with no keyword args
-        ft = xrft.fft(da, real="x", detrend="constant")
+        ft = xrft.fft(da, real_dim="x", detrend="constant")
         # check that the frequency dimension was created properly
         assert ft.dims == ("freq_x",)
         # check that the coords are correct
@@ -230,7 +230,7 @@ class TestfftReal(object):
         npt.assert_allclose(ft_data_expected, ft.values, atol=1e-14)
 
         with pytest.raises(ValueError):
-            xrft.fft(da, real="y", detrend="constant")
+            xrft.fft(da, real_dim="y", detrend="constant")
 
     def test_fft_real_2d(self):
         """
@@ -247,11 +247,11 @@ class TestfftReal(object):
         dx = float(da.x[1] - da.x[0])
         dy = float(da.y[1] - da.y[0])
 
-        daft = xrft.fft(da, real="x")
+        daft = xrft.fft(da, real_dim="x")
         npt.assert_almost_equal(
             daft.values, np.fft.rfftn(da.transpose("y", "x")).transpose()
         )
-        npt.assert_almost_equal(daft.values, xrft.fft(da, dim=["y"], real="x"))
+        npt.assert_almost_equal(daft.values, xrft.fft(da, dim=["y"], real_dim="x"))
 
         actual_freq_x = daft.coords["freq_x"].values
         expected_freq_x = np.fft.rfftfreq(Nx, dx)
@@ -367,7 +367,7 @@ class TestSpectrum(object):
         f_scipy, p_scipy = sps.periodogram(
             da.values, window="rectangular", return_onesided=True
         )
-        ps = xrft.power_spectrum(da, dim="x", real="x", detrend="constant")
+        ps = xrft.power_spectrum(da, dim="x", real_dim="x", detrend="constant")
         npt.assert_almost_equal(ps.values, p_scipy)
 
         A = 20
@@ -427,9 +427,14 @@ class TestSpectrum(object):
         npt.assert_almost_equal(np.ma.masked_invalid(ps).mask.sum(), 0.0)
 
         ps = xrft.power_spectrum(
-            da, dim=["y"], real="x", window="hann", density=False, detrend="constant"
+            da,
+            dim=["y"],
+            real_dim="x",
+            window="hann",
+            density=False,
+            detrend="constant",
         )
-        daft = xrft.fft(da, dim=["y"], real="x", detrend="constant", window="hann")
+        daft = xrft.fft(da, dim=["y"], real_dim="x", detrend="constant", window="hann")
         ps_test = np.real(daft * np.conj(daft))
         f = np.full(ps_test.sizes["freq_x"], 2.0)
         f[0], f[-1] = 1.0, 1.0
@@ -517,12 +522,12 @@ class TestSpectrum(object):
         )
 
         ps = xrft.power_spectrum(
-            da, dim="y", real="x", window="hann", detrend="constant"
+            da, dim="y", real_dim="x", window="hann", detrend="constant"
         )
         npt.assert_array_equal(
             ps.values,
             xrft.power_spectrum(
-                da, dim=["y"], real="x", window="hann", detrend="constant"
+                da, dim=["y"], real_dim="x", window="hann", detrend="constant"
             ).values,
         )
         assert ps.dims == ("time", "freq_y", "freq_x")
@@ -1177,7 +1182,7 @@ def test_real_dft_true_phase():
         },
     )
     s1 = xrft.dft(s, dim="x", true_phase=True, shift=True)
-    s2 = xrft.dft(s, real="x", true_phase=True, shift=True)
+    s2 = xrft.dft(s, real_dim="x", true_phase=True, shift=True)
     s1 = np.conj(s1[{"freq_x": slice(None, s1.sizes["freq_x"] // 2 + 1)}])
     s1 = s1.assign_coords(freq_x=-s1["freq_x"]).sortby("freq_x")
     xrt.assert_allclose(s1, s2)
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -442,7 +442,9 @@ class TestSpectrum(object):
         npt.assert_almost_equal(np.ma.masked_invalid(ps).mask.sum(), 0.0)
 
         with pytest.raises(ValueError):
-            xrft.power_spectrum(da, dim=["y", "x"], window=None, boost_amplitude=True)
+            xrft.power_spectrum(
+                da, dim=["y", "x"], window=None, boost_windowed_amplitude=True
+            )
 
     @pytest.mark.parametrize("dask", [False, True])
     def test_cross_spectrum(self, dask):
@@ -490,7 +492,9 @@ class TestSpectrum(object):
         npt.assert_almost_equal(np.ma.masked_invalid(cs).mask.sum(), 0.0)
 
         with pytest.raises(ValueError):
-            xrft.cross_spectrum(da, da2, dim=dim, window=None, boost_amplitude=True)
+            xrft.cross_spectrum(
+                da, da2, dim=dim, window=None, boost_windowed_amplitude=True
+            )
 
     def test_spectrum_dim(self):
         N = 16
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -384,17 +384,29 @@ class TestSpectrum(object):
             x_da = xr.DataArray(x, coords=[tt], dims=["t"]).chunk({"t": n_segments})
             ps = xrft.xrft.power_spectrum(
                 x_da,
-                dim=["t"],
+                dim="t",
                 window=window_type,
                 chunks_to_segments=True,
                 window_correction=True,
             ).mean("t_segment")
-
+            # Check the energy correction
             npt.assert_allclose(
                 np.sqrt(np.trapz(ps.values, ps.freq_t.values)),
                 A * np.sqrt(2) / 2,
                 rtol=1e-3,
             )
+
+            ps = xrft.xrft.power_spectrum(
+                x_da,
+                dim="t",
+                window=window_type,
+                chunks_to_segments=True,
+                scaling="spectrum",
+                window_correction=True,
+            ).mean("t_segment")
+            # Check the amplitude correction
+            # The factor of 0.5 is there because we're checking the two-sided spectrum
+            npt.assert_allclose(ps.sel(freq_t=fsig), 0.5 * A ** 2 / 2.0)
 
         da = xr.DataArray(
             np.random.rand(2, N, N),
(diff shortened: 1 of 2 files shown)


roxyboy:
I added a flag `correct_amplitude` to account for the amplitude diminishment due to windowing but when you correct for this, the Parseval's identity is no longer satisfied. Maybe @dougiesquire can comment more on this.
I think this has to due to the windowing putting more weighting on lower wavenumbers when the spectrum is not white (because windowing in the spatial domain translates to a convolution in the wavenumber domain) so simply multiplying the mean of the window to the spectra doesn't become equivalent to the variance of original data.

xrft/xrft.py
```diff
@@ -643,14 +635,22 @@ def power_spectrum(da, dim=None, real=None, scaling="density", **kwargs):
...
-    if real is not None:
-        ps = ps * 2
+    for arg in kwargs.keys():
+        if arg == "real" and arg is not None:
+            ps = ps * 2
+        if arg == "window" and correct_amplitude:
+            if arg is not True:
+                raise ValueError("Windowing needs to be turned on.")
+            else:
+                windows, _ = _apply_window(da, dim)
```
roxyboy, line 651: @dougiesquire I think I'm doing this correctly but can you double check?

xrft/tests/test_xrft.py
```diff
@@ -640,10 +648,14 @@ def test_parseval(chunks_to_segments):
         dims=dim,
         coords=da.coords,
     )
+
     # Check that the (rectangular) integral of the spectrum matches the windowed variance
+    # for i in dim:
+    #     da_corrected = (da_prime * window) ** 2 / np.mean(np.hanning(N / n_segments)**2)
```
roxyboy, line 654: @dougiesquire This is where I try to assert the Parseval's identity when correcting for the amplitude change due to windowing. I have commented out for now because I couldn't make it work.

xrft/tests/test_xrft.py
```diff
@@ -640,10 +648,14 @@ def test_parseval(chunks_to_segments):
...
     npt.assert_almost_equal(
         (1 / delta_xy) * ps.mean(fftdim).values,
         ((da_prime * window) ** 2).mean(dim).values,
+        # da_corrected.mean(dim).values,
```
roxyboy, line 658: @dougiesquire When corrected, I think this should assert for `da_corrected`...

dougiesquire (commented):
FYI, I'm going to wait until we've resolved discussions on #104 before reviewing this properly. Just noticed this minor issue while trying to re-familiarise myself with our tests.

navidcy:
@roxyboy I don't agree with having a kwarg "correct_amplitude=True". The amplitude should be corrected always without the user needing to specifically ask for it. Who would like to be returned with the wrong amplitude? Am I missing anything?

dougiesquire:
> @roxyboy I don't agree with having a kwarg eg "correct_amplitude=True". The amplitude should be corrected always without the user needing to specifically ask for it. Who would like to be returned with the wrong amplitude? Am I missing anything?

I agree with this. This is not an option in `scipy`, the correction is just done.

dougiesquire:
The only fiddly thing is that Parseval's may only be approximately satisfied when non-uniform windowing is applied and corrected for. So how to robustly test Parseval's theorem? Is it sufficient to test that we get the same result as `scipy` (see also https://github.com/xgcm/xrft/issues/104#issuecomment-774572393)?

navidcy:
> > @roxyboy I don't agree with having a kwarg eg "correct_amplitude=True". The amplitude should be corrected always without the user needing to specifically ask for it. Who would like to be returned with the wrong amplitude? Am I missing anything?
> 
> I agree with this. This is not an option in `scipy`, the correction is just done.
> 
> It's also not strictly the amplitude that we are correcting for, but the energy - see https://community.sw.siemens.com/s/article/window-correction-factors.

Continuing this thread: what does the `idft()` function expect then `real=True`? Does it expect to get a data array with the *correct* amplitude or with the *wrong/halved* amplitude? It should be the former, right?

dougiesquire:
> > > @roxyboy I don't agree with having a kwarg eg "correct_amplitude=True". The amplitude should be corrected always without the user needing to specifically ask for it. Who would like to be returned with the wrong amplitude? Am I missing anything?
> > 
> > 
> > I agree with this. This is not an option in `scipy`, the correction is just done.
> > It's also not strictly the amplitude that we are correcting for, but the energy - see https://community.sw.siemens.com/s/article/window-correction-factors.
> 
> Continuing this thread: what does the `idft()` function expect then `real=True`? Does it expect to get a data array with the _correct_ amplitude or with the _wrong/halved_ amplitude? It should be the former, right?

Sorry, I think we could be talking about two different issues that are being addressed in this PR:
1. Using a `real` dimension returns a one-sided spectrum that has not had its energy compensated for being one-sided and so integrates to only half of the signal energy (#105).
2. Applying non-uniform weighting reduces the energy in a signal and we currently don't correct for this (#104).

Regarding 1., I think the correction (essentially a multiplication by 2) should happen by default. But this should only be applied within the `*_spectrum` functions. `dft()` and `idft()` need not be modified.

Regarding 2., I also think the correction (scaling by the energy of the window) should happen by default. But we need to think about how to test this as this correction may only approximately satisfy Parseval's theorem (still better than not correcting, which will significantly underestimate energy).

roxyboy:
When starting off when spectrum, one of the first things a user usually tests is to see if one is doing it correctly is to see whether the Parseval's identity is satisfied. Therefore, I find the correction only approximately satisfying the Parseval's identity quite troubling...

roxyboy:
> Regarding 1., I think the correction (essentially a multiplication by 2) should happen by default. But this should only be applied within the `*_spectrum` functions. `dft()` and `idft()` need not be modified.

This is already done in this PR.

roxyboy:
> I therefore, vote for keeping the default behavior as not correcting for the amplitude and returning an output that satisfies Parseval's exactly.

In the current version of `xrft`, the Parseval's identity is satisfied exactly for the spatial variance of the windowed data. https://github.com/xgcm/xrft/blob/7381a801edb3605db123dbfffde840a1f2097df4/xrft/tests/test_xrft.py#L633

dougiesquire:
> > Regarding 2., I also think the correction (scaling by the energy of the window) should happen by default. But we need to think about how to test this as this correction may only approximately satisfy Parseval's theorem (still better than not correcting, which will significantly underestimate energy).
> 
> When starting off when spectrum, one of the first things a user (and also I) usually test is to see if one is doing it correctly is to check whether the Parseval's identity is satisfied. Therefore, I find the correction only approximately satisfying the Parseval's identity quite troubling...
> I therefore, vote for keeping the default behavior as not correcting for the amplitude and returning an output that satisfies Parseval's exactly.

I disagree. The test that you're referring to as "satisfying Parseval's exactly" actually tests that the integral of the windowed spectrum matches the energy of the _windowed_ signal. I don't think this is a very meaningful test. I think that most users will expect to be able to calculate the psd/csd and interpret the magnitude of the output in the same way regardless of the windowing applied - that is, as an indication of the power as a function of frequency. That is not the current behaviour. Right now, the magnitude of a spectrum is substantially attenuated if `window=True`.

Just to be clear, `scipy` spectral functions apply this correction - there's not even an option to not apply it.

roxyboy:
> Just to be clear, `scipy` spectral functions apply this correction - there's not even an option to not apply it.

Right, and I would probably be confused as to why `scipy` didn't satisfy the Parseval's identity if it were have not been for this discussion.

roxyboy:
And I'm not sure what you mean by it not being a meaningful test... Based on Parseval's theorem, `the integral of the windowed spectrum matches the energy of the windowed signal` is indeed what should happen. It's one of the fundamental justifications why one can use spectrum as a method to decompose the energy into wavenumber modes.

dougiesquire:
> And I'm not sure what you mean by it not being a meaningful test... Based on Parseval's theorem, `the integral of the windowed spectrum matches the energy of the windowed signal` is indeed what should happen. It's one of the fundamental justifications why one can use spectrum as a method to decompose the energy into wavenumber modes.

I agree that this test must pass if one hasn't corrected for the energy reduction due to windowing. But I personally don't think an uncorrected spectra is very useful. The window is only there to help reduce leakage during the FFT so I think its influence on the power should be accounted for. I'm not sure why anyone would want to compute a power spectrum that doesn't give a direct indication of the power in the input signal.

The whole point in the correction is so that Parseval's theorem is (approximately) satisfied. Without the correction, the integral of a windowed spectrum is reduced relative to the energy of the input signal. This isn't a great webpage, but it might be useful: https://community.sw.siemens.com/s/article/window-correction-factors.

roxyboy:
> I agree that this test must pass if one hasn't corrected for the energy reduction due to windowing. But I personally don't think an uncorrected spectra is very useful. The window is only there to help reduce leakage during the FFT so I think its influence on the power should be accounted for. I'm not sure why anyone would want to compute a power spectrum that doesn't give a direct indication of the power in the input signal.

I can only speak for my personal experience but I was only interested in the spectral slopes and seasonal change in the energy and not the absolute energy itself so for me, the windowed data satisfying the Parseval's theorem was the priority. But I also see your point so that is why I think having the option as the `boost_amplitude` is the best way to go.

dougiesquire:
> But I also see your point so that is why I think having the option as the `boost_amplitude` is the best way to go.

What about `correct_energy`? Whatever you decide, as always, clear documentation is probably what's really crucial. Let me know if/how I can help out with this.

lanougue:
@roxyboy Just a comment here. I might be wrong but you multiplied by 2 the *_spectrum() outputs when a "real" is in args. However, first value (null frequency) and last value (highest frequency) should not be multriplied by two. (Same in scipy). See first figure of @105.
I think that line cs = cs * 2 and ps = ps*2 should be complemented by something like:
```
cs[{real:0}] = 0.5*a[{real:0}]
cs[{real:-1}] = 0.5*a[{real:-1}]
```

xrft/tests/test_xrft.py
```diff
@@ -352,15 +352,33 @@ class TestSpectrum(object):
...
+        f_scipy, p_scipy = sps.periodogram(da.values,
+                                           window='rectangular',
+                                           return_onesided=True
+                                           )
+        ps = xrft.xrft.power_spectrum(da,
+                                      dim='x',
+                                      real='x',
+                                      detrend='constant'
+                                      )
+        npt.assert_almost_equal(ps.values[1:-1], p_scipy[1:-1])
```
lanougue, line 368: First and last values should match too between scipy and xrft.power_spectrum.
These two values should not be multiplied by 2 in xrft.power_spectrum. See my comment in the current discussion

rabernat:
> I think that line cs = cs * 2 and ps = ps*2 should be complemented by something like:
> 
> ```
> cs[{real:0}] = 0.5*a[{real:0}]
> cs[{real:-1}] = 0.5*a[{real:-1}]
> ```

This will not work for dask-backed arrays because you can't assign to them. So we would need a lazy version of this function.

lanougue:
The expected behaviour of returning single sided spectrum is to discard redundant information. Null and highest frequencies are NOT duplicated in two-sided spectrum. Thus, they should not be doubled. However, all the other frequencies are duplicated. They just can be safely discarded (and doubled if we choose scipy convention)

lanougue:
> > I think that line cs = cs * 2 and ps = ps*2 should be complemented by something like:
> > ```
> > cs[{real:0}] = 0.5*a[{real:0}]
> > cs[{real:-1}] = 0.5*a[{real:-1}]
> > ```
> 
> This will not work for dask-backed arrays because you can't assign to them. So we would need a lazy version of this function.

Ok, I suggest to remove cs = cs * 2 and do:
```
f = np.full(cs.sizes[real], 2., dtype=cs.dtype)
f[0],f[-1]=1.,1.
cs = cs * xr.DataArray(f, dims='x')
```

roxyboy:
> Ok, I suggest to remove cs = cs * 2 and do:
> 
> ```
> f = np.full(cs.sizes[real], 2.)
> f[0],f[-1]=1.,1.
> cs = cs * xr.DataArray(f, dims=real)
> ```

Done :)

roxyboy:
> What about `correct_energy`? Whatever you decide, as always, clear documentation is probably what's really crucial. Let me know if/how I can help out with this.

Cross spectrum isn't really "energy" so I think `boost_amplitude` is better...

xrft/xrft.py
```diff
@@ -643,7 +643,9 @@ def power_spectrum(da, dim=None, scaling="density", boost_amplitude=False, **kwa
 
     for arg in kwargs.keys():
         if arg == "real" and kwargs[arg] is not None:
-            ps = ps * 2
+            f = np.full(ps.sizes["freq_" + kwargs[arg]], 2.0)
+            f[0], f[-1] = 1.0, 1.0
+            ps = ps * xr.DataArray(f, dims="freq_" + kwargs[arg])
```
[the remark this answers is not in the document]
lanougue: I would code:
```
real_dim =[d for d in updated_dims if real==d[-len(real):]][0] # find transformed real dimension
f = np.full(ps.sizes[real_dim], 2.0)
f[0], f[-1] = 1.0, 1.0
ps = ps * xr.DataArray(f, dims=real_dim)
```

dougiesquire:
> > > I think that line cs = cs * 2 and ps = ps*2 should be complemented by something like:
> > > ```
> > > cs[{real:0}] = 0.5*a[{real:0}]
> > > cs[{real:-1}] = 0.5*a[{real:-1}]
> > > ```
> > 
> > 
> > This will not work for dask-backed arrays because you can't assign to them. So we would need a lazy version of this function.
> 
> Ok, I suggest to remove cs = cs * 2 and do:
> 
> ```
> f = np.full(cs.sizes[real], 2.)
> f[0],f[-1]=1.,1.
> cs = cs * xr.DataArray(f, dims=real)
> ```

I think the behaviour depends on whether the number of points is even or odd (see [numpy.fft.rfft](https://numpy.org/doc/stable/reference/generated/numpy.fft.rfft.html))
For even inputs, we would want something like what @lanougue has above, but for odd inputs, I don't think the last (highest) frequency should be replaced with 1., e.g.
```
f = np.full(cs.sizes[real], 2.)
f[0]=1.
cs = cs * xr.DataArray(f, dims=real)
```

navidcy:
> > > > I think that line cs = cs * 2 and ps = ps*2 should be complemented by something like:
> > > > ```
> > > > cs[{real:0}] = 0.5*a[{real:0}]
> > > > cs[{real:-1}] = 0.5*a[{real:-1}]
> > > > ```
> > > 
> > > 
> > > This will not work for dask-backed arrays because you can't assign to them. So we would need a lazy version of this function.
> > 
> > 
> > Ok, I suggest to remove cs = cs * 2 and do:
> > ```
> > f = np.full(cs.sizes[real], 2.)
> > f[0],f[-1]=1.,1.
> > cs = cs * xr.DataArray(f, dims=real)
> > ```
> 
> I think the behaviour depends on whether the number of points is even or odd (see [numpy.fft.rfft](https://numpy.org/doc/stable/reference/generated/numpy.fft.rfft.html))
> For even inputs, we would want something like what @lanougue has above, but for odd inputs, I don't think the last (highest) frequency should be replaced with 1., e.g.
> 
> ```
> f = np.full(cs.sizes[real], 2.)
> f[0]=1.
> cs = cs * xr.DataArray(f, dims=real)
> ```

Nice catch! (In an ideal world, a test with odd number of points should have raised a red flag. :) )

lanougue:
@roxyboy 
real flag is finally quite important in how power_spectrum() behave. I think it could be added in the signature of the function.
It would also avoid all kwargs[arg] everywhere.
I would suggest to just adding
`if real is not None:...`

xrft/tests/test_xrft.py
```diff
@@ -121,9 +121,10 @@ def test_fft_2d(self):
         ft = xrft.fft(da, shift=False)
         npt.assert_almost_equal(ft.values, np.fft.fftn(da.values))
 
-        ft = xrft.fft(da, shift=False, window=True, detrend="constant")
+        ft = xrft.fft(da, shift=False, window="hann", detrend="constant")
         dim = da.dims
-        window = np.hanning(N) * np.hanning(N)[:, np.newaxis]
+        # window = np.hanning(N) * np.hanning(N)[:, np.newaxis]
```
navidcy, line 126: remove commented code?

xrft/xrft.py
```diff
@@ -933,8 +986,8 @@ def isotropic_power_spectrum(
     density : list, optional
         If true, it will normalize the spectrum to spectral density
     window : bool, optional
-        Whether to apply a Hann window to the data before the Fourier
-        transform is taken
+        Whether to apply a window to the data before the Fourier
+        transform is taken. Please adhere to scipy.signal.windows' naming convention.
```
navidcy, line 990: suggest: Please adhere to `scipy.signal.windows` for naming convention.

xrft/xrft.py
```diff
@@ -1022,8 +1075,8 @@ def isotropic_cross_spectrum(
     density : list (optional)
         If true, it will normalize the spectrum to spectral density
     window : bool (optional)
-        Whether to apply a Hann window to the data before the Fourier
-        transform is taken
+        Whether to apply a window to the data before the Fourier
+        transform is taken. Please adhere to scipy.signal.windows' naming convention.
```
navidcy, line 1079: same here; I suggest: Please adhere to scipy.signal.windows for naming convention.

navidcy (changes requested):
Some very minor suggestions..

dougiesquire:
> Cross spectrum isn't really "energy" so I think `boost_amplitude` is better...

My only concern with having "amplitude" in the name is that there is a _different_ correction that specifically corrects the amplitude of the spectrum (rather than its integral/energy). So I worry that `boost_amplitude` could be confusing for users.

lanougue:
It could be easy to change window ampltude in order to have the windowed signal to have the same energy (variance) as the original signal. Thus, paserval will be perfectly respected in both cases (windowed and not windowed).
The user will thus have no need to worry about it

roxyboy:
> It could be easy to change window ampltude in order to have the windowed signal to have the same energy (variance) as the original signal. Thus, paserval will be perfectly respected in both cases (windowed and not windowed).
> The user will thus have no need to worry about it

I'm not sure how to do this consistently. Do you have example code?

lanougue:
> > It could be easy to change window ampltude in order to have the windowed signal to have the same energy (variance) as the original signal. Thus, paserval will be perfectly respected in both cases (windowed and not windowed).
> > The user will thus have no need to worry about it
> 
> I'm not sure how to do this consistently. Do you have example code?

I edited my comment to explain. What do you think ?

dougiesquire:
> > > It could be easy to change window ampltude in order to have the windowed signal to have the same energy (variance) as the original signal. Thus, paserval will be perfectly respected in both cases (windowed and not windowed).
> > > The user will thus have no need to worry about it
> > 
> > 
> > I'm not sure how to do this consistently. Do you have example code?
> 
> I edited my comment to explain. What do you think ?

Do you just mean to scale the windowed spectrum by the ratio of the windowed and non-windowed rms (rather than by the correction factor computed from the window) - i.e. enforce Parseval's?

dougiesquire:
> I edited my comment to explain. What do you think ?

Sure we could do this. In my very limited experience this is not commonly done but I don't know why not...

dougiesquire:
I've just noticed that `scipy` actually has a test for the window correction using a simple sinusoidal signal. `xrft` obviously also passes this test if the original correction I suggested is applied. Run-able code below.

```python
import xrft
import scipy.signal as sps
from numpy.testing import assert_allclose

# The following is copied from scipy test function test_window_correction() 
# see https://github.com/scipy/scipy/blob/master/scipy/signal/tests/test_spectral.py#L485
# 
# I've made a few minor edits so we can easily run the same test using xrft:
#   1. Added return_onesided=False
#   2. Added noverlap=0 
#   3. Test the 'Hann' window only
#   4. Commented out the tests of scaling='spectrum'

A = 20
fs = 1e4
nperseg = int(fs//10)
fsig = 300
ii = int(fsig*nperseg//fs)  # Freq index of fsig

tt = np.arange(fs)/fs
x = A*np.sin(2*np.pi*fsig*tt)

for window in ['hann']: #, 'bartlett', ('tukey', 0.1), 'flattop']:
    # _, p_spec = sps.welch(x, fs=fs, nperseg=nperseg, window=window, 
    #                       noverlap=0, return_onesided=False,
    #                       scaling='spectrum')
    freq, p_dens = sps.welch(x, fs=fs, nperseg=nperseg, window=window, 
                             noverlap=0, return_onesided=False,
                             scaling='density')

    # Check peak height at signal frequency for 'spectrum'
    # assert_allclose(p_spec[ii], A**2/2.0)
    # Check integrated spectrum RMS for 'density'
    assert_allclose(np.sqrt(np.trapz(p_dens, freq)), A*np.sqrt(2)/2,
                    rtol=1e-3)
```
And using `xrft`:
```python
x_da = xr.DataArray(x, coords=[tt], dims=['t']).chunk({'t': nperseg})

p_xrft = xrft.xrft.power_spectrum(x_da, 
                                  dim=['t'],
                                  window=True,
                                  chunks_to_segments=True).mean('t_segment')

corr = 1 / np.mean(sps.hann(nperseg, sym=False) ** 2)

assert_allclose(np.sqrt(np.trapz(corr * p_xrft.values, p_xrft.freq_t.values)), A*np.sqrt(2)/2,
                rtol=1e-3)
```

roxyboy:
> My only concern with having "amplitude" in the name is that there is a _different_ correction that specifically corrects the amplitude of the spectrum (rather than its integral/energy). So I worry that `boost_amplitude` could be confusing for users.

I changed it to `boost_windowed_amplitude`.

roxyboy:
Ok, I've made all the changes suggested. I think this is ready to merge. @rabernat ??

roxyboy:
> And using `xrft`:
> 
> ```python
> x_da = xr.DataArray(x, coords=[tt], dims=['t']).chunk({'t': nperseg})
> 
> p_xrft = xrft.xrft.power_spectrum(x_da, 
>                                   dim=['t'],
>                                   window=True,
>                                   chunks_to_segments=True).mean('t_segment')
> 
> corr = 1 / np.mean(sps.hann(nperseg, sym=False) ** 2)
> 
> assert_allclose(np.sqrt(np.trapz(corr * p_xrft.values, p_xrft.freq_t.values)), A*np.sqrt(2)/2,
>                 rtol=1e-3)
> ```

Actually, this test is supposed to hold for `boost_windowed_amplitude = True`??

dougiesquire:
> > And using `xrft`:
> > 
> > ```python
> > x_da = xr.DataArray(x, coords=[tt], dims=['t']).chunk({'t': nperseg})
> > 
> > p_xrft = xrft.xrft.power_spectrum(x_da, 
> >                                   dim=['t'],
> >                                   window=True,
> >                                   chunks_to_segments=True).mean('t_segment')
> > 
> > corr = 1 / np.mean(sps.hann(nperseg, sym=False) ** 2)
> > 
> > assert_allclose(np.sqrt(np.trapz(corr * p_xrft.values, p_xrft.freq_t.values)), A*np.sqrt(2)/2,
> >                 rtol=1e-3)
> > ```
> 
> Actually, this test is supposed to hold for `boost_windowed_amplitude = True`?? As of now, it seems to be passing for `boost_windowed_amplitude = False`...

Do you mean in your implementation or in the above? In the above, I have applied the correction with the `corr` variable. Without this, the test does not pass.

roxyboy:
> Do you mean in your implementation or in the above? In the above, I have applied the correction with the `corr` variable. Without this, the test does not pass.

Makes sense. Now the test is passing with `boost_windowed_amplitude = True` without the `corr` factor :)

xrft/xrft.py
```diff
@@ -719,18 +769,34 @@ def cross_spectrum(da1, da2, dim=None, scaling="density", **kwargs):
...
+            f[0] = 1.0
+        cs = cs * xr.DataArray(f, dims=real_dim)
+
+    for arg in kwargs.keys():
+        if arg == "window" and boost_windowed_amplitude:
+            if kwargs[arg] is None:
+                raise ValueError("Windowing needs to be turned on.")
+            else:
+                windows, _ = _apply_window(da, dim, window_type=kwargs[arg])
+                cs = cs / (windows ** 2).mean()
```
dougiesquire, line 790: As above

dougiesquire (commented):
Thanks for doing this @roxyboy. I've made a few minor comments.

I still think that `boost_windowed_amplitude` could be a confusing name for reasons I've outlined above. But, if other folks are happy with this name then I am too :)

xrft/xrft.py
```diff
@@ -641,30 +662,53 @@ def power_spectrum(da, dim=None, scaling="density", **kwargs):
...
+            f[0] = 1.0
+        ps = ps * xr.DataArray(f, dims=real_dim)
+
+    for arg in kwargs.keys():
+        if arg == "window" and boost_windowed_amplitude:
+            if kwargs[arg] is None:
+                raise ValueError("Windowing needs to be turned on.")
+            else:
+                windows, _ = _apply_window(da, dim, window_type=kwargs[arg])
+                ps = ps / (windows ** 2).mean()
```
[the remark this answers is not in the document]
dougiesquire: UPDATE: I've made a little progress in my understanding here. I'll provide an update on the main thread and see what folks think for moving forward.

rabernat:
I appreciate everyone's hard work on this. Sorry I haven't had more time to engage. I will try to review within a few days.

dougiesquire:
I've done a little more investigation and this is where I'm at:

Applying a non-uniform windowing weighting to a signal reduces its energy. There are two very commonly applied corrections for this:
1. __Amplitude correction__ - this corrects the amplitude of peaks in the spectrum. It is typically applied to the power spectrum (ie not density) and is most useful in strongly periodic signals. It ensures, for example, that the peak in the power spectrum of a 10 Hz sine wave with RMS=1.0 has a magnitude of 1.0. In scipy, this correction is applied by default when `scaling='spectrum'` without any option to turn it off.
2. __Energy correction__ - this corrects for the energy (integral) of the spectrum. It is typically applied to the power spectral density (including cross power spectral density). It ensures that the integral of the spectral density (approximately) matches the square of the RMS of the signal. In scipy, this correction is applied by default when `scaling='density'` without any option to turn it off.

The energy correction is only exact when the signal being windowed is independent of the window. It gets more accurate as the width of the window gets large in comparison with any noticeable period in the signal (see discussion here: https://dsp.stackexchange.com/questions/47598/does-windowing-affect-parsevals-theorem). This is why we can't exactly satisfy Parseval's theorem with the energy correction.

Moving forward, I propose that we add an option (e.g. `window_correction`) that matches the behaviour of scipy. That is:
- With `window_correction = True` and `scaling='density'`, the energy correction is applied
- With `window_correction = True` and `scaling='spectrum'`, the amplitude correction is applied.

I'll write out below specifically how these are applied and how we can test their implementation.

What became very clear to me while trying to read about this is that there is a lot of confusion around these corrections and whether/how they're applied in a variety of software. So, whatever we decide to do, let's try and document it very clearly!

xrft/xrft.py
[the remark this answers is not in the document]
roxyboy: The correction applied in scipy is `.sum()` while I've currently implemented the correction as `.mean()`. We should apply the correction with `.sum()`...?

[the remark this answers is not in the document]
dougiesquire: `mean()` is correct with the current implementation.

This is because scaling by the sampling frequency (or it's square if `scaling='spectrum'`) is done as a separate step in xrft, whereas scipy does all the scaling in one hit with the `scale` variable (https://github.com/scipy/scipy/blob/master/scipy/signal/spectral.py#L1798). It's simple to show that by using the `mean()` instead of the `sum()` to compute our corrections, we end up with the same final scaling as scipy.

dougiesquire:
Sorry, it took me a while to get to this. I think @roxyboy has already added some of what I summarise below. But just documenting here for clarity.
 
With the scaling that is already applied to the spectrum within xrft, the correction factors are computed as follows:
 
__Energy correction__:
```python
correction = 1.0 / (window ** 2).mean()
```
__Amplitude correction__:
```python
correction = 1.0 / window.mean() ** 2
```
 
With these applied when `scaling='density'` and `'spectrum'`, respectively, xrft will give the same results as scipy.
 
We can test these by borrowing scipy's test function:
```python
A = 20
fs = 1e4
n_segments = int(fs // 10)
fsig = 300

tt = np.arange(fs) / fs
x = A * np.sin(2 * np.pi * fsig * tt)

for window_type in ["hann", "bartlett", "tukey", "flattop"]:
    # See https://github.com/scipy/scipy/blob/master/scipy/signal/tests/test_spectral.py#L485

    x_da = xr.DataArray(x, coords=[tt], dims=["t"]).chunk({"t": n_segments})
    
    p_dens = xrft.xrft.power_spectrum(
        x_da,
        dim=["t"],
        window=window_type,
        chunks_to_segments=True,
        scaling='density',
        window_correction=True,
    ).mean("t_segment")
    
    p_spec = xrft.xrft.power_spectrum(
        x_da,
        dim=["t"],
        window=window_type,
        chunks_to_segments=True,
        scaling='spectrum',
        window_correction=True,
    ).mean("t_segment")

    # Check the energy correction for p_dens
    npt.assert_allclose(
        np.sqrt(np.trapz(p_dens.values, p_dens.freq_t.values)),
        A * np.sqrt(2) / 2,
        rtol=1e-3)
    
    # Check the amplitude correction for p_spec
    # The factor of 0.5 is there because we're checking the two-sided spectrum
    npt.assert_allclose(
        p_spec.sel(freq_t=fsig),
        0.5 * A ** 2 / 2.0) 
```

roxyboy:
> Sorry, it took me a while to get to this. I think @roxyboy has already added some of what I summarise below. But just documenting here for clarity.
> 
> With the scaling that is already applied to the spectrum within xrft, the correction factors are computed as follows:
> 
> **Energy correction**:
> 
> ```python
> correction = 1.0 / (window ** 2).mean()
> ```
> 
> **Amplitude correction**:
> 
> ```python
> correction = 1.0 / window.mean() ** 2
> ```
> 
> With these applied when `scaling='density'` and `'spectrum'`, respectively, xrft will give the same results as scipy.
> 
> We can test these by borrowing scipy's test function:
> 
> ```python
> A = 20
> fs = 1e4
> n_segments = int(fs // 10)
> fsig = 300
> 
> tt = np.arange(fs) / fs
> x = A * np.sin(2 * np.pi * fsig * tt)
> 
> for window_type in ["hann", "bartlett", "tukey", "flattop"]:
>     # See https://github.com/scipy/scipy/blob/master/scipy/signal/tests/test_spectral.py#L485
> 
>     x_da = xr.DataArray(x, coords=[tt], dims=["t"]).chunk({"t": n_segments})
>     
>     p_dens = xrft.xrft.power_spectrum(
>         x_da,
>         dim=["t"],
>         window=window_type,
>         chunks_to_segments=True,
>         scaling='density',
>         window_correction=True,
>     ).mean("t_segment")
>     
>     p_spec = xrft.xrft.power_spectrum(
>         x_da,
>         dim=["t"],
>         window=window_type,
>         chunks_to_segments=True,
>         scaling='spectrum',
>         window_correction=True,
>     ).mean("t_segment")
> 
>     # Check the energy correction for p_dens
>     npt.assert_allclose(
>         np.sqrt(np.trapz(p_dens.values, p_dens.freq_t.values)),
>         A * np.sqrt(2) / 2,
>         rtol=1e-3)
>     
>     # Check the amplitude correction for p_spec
>     # The factor of 0.5 is there because we're checking the two-sided spectrum
>     npt.assert_allclose(
>         p_spec.sel(freq_t=fsig),
>         0.5 * A ** 2 / 2.0) 
> ```

All done :)

xrft/xrft.py
```diff
@@ -641,30 +663,65 @@ def power_spectrum(da, dim=None, scaling="density", **kwargs):
...
+        else:
+            f[0] = 1.0
+        ps = ps * xr.DataArray(f, dims=real_dim)
+
     if scaling == "density":
+        if window_correction:
+            for arg in kwargs.keys():
+                if arg == "window":
+                    if kwargs[arg] is None:
+                        raise ValueError("Windowing needs to be turned on.")
```
dougiesquire, line 688: Maybe a little more verbose, e.g
"window_correction can only be applied when a non-uniform window is applied"
or, this could just silently do nothing?

xrft/xrft.py
```diff
@@ -641,30 +663,65 @@ def power_spectrum(da, dim=None, scaling="density", **kwargs):
...
+                    else:
+                        msg = "Energy correction is applied where the integral of the spectral density (approximately) matches the square of the RMS of the input signal."
+                        warnings.warn(msg, Warning)
```
dougiesquire, line 691: This could get quite annoying. Do we really need a warning every time the user chooses to use this parameter?

xrft/xrft.py
```diff
@@ -641,30 +663,65 @@ def power_spectrum(da, dim=None, scaling="density", **kwargs):
...
+                        windows, _ = _apply_window(da, dim, window_type=kwargs[arg])
+                        ps = ps / (windows ** 2).mean()
         fs = np.prod([float(ps[d].spacing) for d in updated_dims])
         ps *= fs
     elif scaling == "spectrum":
+        if window_correction:
+            for arg in kwargs.keys():
+                if arg == "window":
+                    if kwargs[arg] is None:
```
dougiesquire, line 700: Same comments as above

xrft/xrft.py
```diff
@@ -682,9 +739,16 @@ def cross_spectrum(da1, da2, dim=None, scaling="density", **kwargs):
     dim : str or sequence of str, optional
         The dimensions along which to take the transformation. If `None`, all
         dimensions will be transformed.
+    real : str, optional
+        Real Fourier transform will be taken along this dimension.
     scaling : str, optional
         If 'density', it will normalize the output to power spectral density
         If 'spectrum', it will normalize the output to power spectrum
+    window_correction : boolean
+        If True, it will correct for the energy reduction resulting from applying a non-uniform window.
```
dougiesquire, line 748: As above

xrft/xrft.py
```diff
@@ -719,18 +783,46 @@ def cross_spectrum(da1, da2, dim=None, scaling="density", **kwargs):
...
+        cs = cs * xr.DataArray(f, dims=real_dim)
+
     if scaling == "density":
+        if window_correction:
+            for arg in kwargs.keys():
+                if arg == "window":
+                    if kwargs[arg] is None:
+                        raise ValueError("Windowing needs to be turned on.")
+                    else:
+                        msg = "Energy correction is applied where the integral of the spectral density (approximately) matches the square of the RMS of the input signal."
```
dougiesquire, line 805: As above

xrft/xrft.py
```diff
@@ -719,18 +783,46 @@ def cross_spectrum(da1, da2, dim=None, scaling="density", **kwargs):
...
         fs = np.prod([float(cs[d].spacing) for d in updated_dims])
         cs *= fs
     elif scaling == "spectrum":
+        if window_correction:
+            for arg in kwargs.keys():
+                if arg == "window":
+                    if kwargs[arg] is None:
+                        raise ValueError("Windowing needs to be turned on.")
+                    else:
+                        msg = "Amplitude correction is applied which corrects the amplitude of peaks in the spectrum."
```
dougiesquire, line 818: As above

dougiesquire (commented):
Crikey you're efficient @roxyboy. Making changes as I review!

All looks good to me I think. Thanks for persevering here. I think this is a useful addition and it's great to also now have the option to use all the scipy window functions.

xrft/xrft.py
roxyboy: I personally would... but that's only my opinion.

roxyboy: I personally would... but that's only my opinion. Perhaps others could chime in @navidcy, @rabernat, @lanougue

xrft/xrft.py
roxyboy: Silently doing nothing is definitely going to confuse users.

xrft/xrft.py
roxyboy: Users can suppress warnings when they use any package fwiw.

lanougue:
Sorry for not being of any help here. I am in the middle of a satellite cal/Val campaign. On a boat with stomach upside down....

roxyboy:
@lanougue Good luck on the cruise! Cal/Val for SWOT??

lanougue:
> @lanougue Good luck on the cruise! Cal/Val for SWOT??
Thanks @roxyboy,
One month cruise for CFOSAT this time... Maybe later for SWOT.

rabernat:
Is this ready for me to review now?

roxyboy:
> Is this ready for me to review now?

@rabernat I believe so. I don't intend to make further edits for this PR.

xrft/xrft.py
```diff
@@ -933,8 +1037,8 @@ def isotropic_power_spectrum(
     density : list, optional
         If true, it will normalize the spectrum to spectral density
     window : bool, optional
-        Whether to apply a Hann window to the data before the Fourier
-        transform is taken
+        Whether to apply a window to the data before the Fourier
+        transform is taken. Please adhere to scipy.signal.windows for naming convention.
```
rabernat, line 1041: Is `window` allowed to be a string (e.g. `window="Hann"`)? If so, the docstring needs an update (change `bool`).

xrft/xrft.py
```diff
@@ -1022,8 +1126,8 @@ def isotropic_cross_spectrum(
     density : list (optional)
         If true, it will normalize the spectrum to spectral density
     window : bool (optional)
-        Whether to apply a Hann window to the data before the Fourier
-        transform is taken
+        Whether to apply a window to the data before the Fourier
+        transform is taken. Please adhere to scipy.signal.windows for naming convention.
```
rabernat, line 1130: Same.

xrft/xrft.py
```diff
@@ -719,18 +791,50 @@ def cross_spectrum(da1, da2, dim=None, scaling="density", **kwargs):
...
+        if len(da1[real]) % 2 == 0:
+            f[0], f[-1] = 1.0, 1.0
+        else:
+            f[0] = 1.0
+        cs = cs * xr.DataArray(f, dims=real_dim)
+
     if scaling == "density":
+        if window_correction:
+            for arg in kwargs.keys():
+                if arg == "window":
```
rabernat, line 809: Why do we need to iterate through the args here?

Why not `window = kwargs.get("window", None)`?

That would let you reduce the level of nesting.

xrft/xrft.py
rabernat: I'm not sure I understand what to do with this warning. It seems like a simple description of what the option does. That belongs in the documentation. Why does that merit a warning? How should a user respond when they see it? What action do you hope to provoke from the user?

I suspect that the warning is here because some users may find this behavior unexpected. (But then why would they ask for window_correction=True?) Is there an alternative behavior they can turn on? How?

In summary, if we use a warning, let's try to have the warning suggest a concrete action the user can take to mitigate confusion or avoid incorrect / unwanted results. The warning should not appear every time a user uses a specific option, unless that option is deprecated.

xrft/xrft.py
```diff
@@ -641,30 +665,69 @@ def power_spectrum(da, dim=None, scaling="density", **kwargs):
...
     elif scaling == "spectrum":
+        if window_correction:
+            for arg in kwargs.keys():
+                if arg == "window":
+                    if kwargs[arg] is None:
+                        raise ValueError(
+                            "window_correction can only be applied when windowing is turned on."
+                        )
+                    else:
+                        msg = "Amplitude correction is applied which corrects the amplitude of peaks in the spectrum."
```
rabernat, line 709: same comment about warnings.

xrft/xrft.py
```diff
@@ -41,29 +41,59 @@ def _fft_module(da):
...
+        "flattop",
+        "gaussian",
+        "general_cosine",
+        "general_gaussian",
+        "general_hamming",
+        "triang",
+        "nuttall",
+    ]:
+        raise NotImplementedError(
+            "Please adhere to scipy.signal.windows for naming convention."
```
rabernat, line 73:
```suggestion
            f"Window type {window_type} not supported. Please adhere to scipy.signal.windows for naming convention."
```

xrft/xrft.py
```diff
@@ -641,30 +665,69 @@ def power_spectrum(da, dim=None, scaling="density", **kwargs):
...
+    if real is not None:
+        real_dim = [d for d in updated_dims if real == d[-len(real) :]][
+            0
+        ]  # find transformed real dimension
+        f = np.full(ps.sizes[real_dim], 2.0)
+        if len(da[real]) % 2 == 0:
+            f[0], f[-1] = 1.0, 1.0
+        else:
+            f[0] = 1.0
+        ps = ps * xr.DataArray(f, dims=real_dim)
```
rabernat, line 683: Are we sure this multiplication works if `ps` has coordinates?

xrft/xrft.py
```diff
@@ -621,9 +636,18 @@ def power_spectrum(da, dim=None, scaling="density", **kwargs):
     dim : str or sequence of str, optional
         The dimensions along which to take the transformation. If `None`, all
         dimensions will be transformed.
+    real : str, optional
```
rabernat, line 639:
```suggestion
    real_dim : str, optional
```

I think it would be easier to understand this argument if it were explicitly named `real_dim`.

rabernat (changes requested):
Thanks so much for everyone's work on this! Overall it seems great.

My main concern is that we are trying to use warnings where we should be using documentation. I don't think this PR is complete without __narrative documentation__ that fully explains all of the different choices for `window`, `scaling`, `wind_corrections`, `real` etc. These options are too complex to expect people will figure them out via a doc string.

Is it safe to assume that these examples will be part of #135? If so, then we can skip that here. But let's recognize that they are closely coupled.

roxyboy:
> My main concern is that we are trying to use warnings where we should be using documentation. I don't think this PR is complete without **narrative documentation** that fully explains all of the different choices for `window`, `scaling`, `wind_corrections`, `real` etc. These options are too complex to expect people will figure them out via a doc string.
> 
> Is it safe to assume that these examples will be part of #135? If so, then we can skip that here. But let's recognize that they are closely coupled.

I 100% agree with this and a better documentation is on my to-do list as part of PR #135 .

roxyboy:
Well, this is annoying... The tests are passing locally...

roxyboy:
@rabernat I implemented your changes :) Can we merge this?

navidcy:
I agree with @rabernat that serious update of docs is in order. But given that this will happen in a sister-PR soon then I'm also in favor of going forward with this.

(I'm happy to help with the docs-PR, but somebody needs to show me how to compile/view docs locally :))

navidcy:
Also note that there are breaking changes in this PR (e.g., `real`->`real_dim`) so a new minor release should follow (after doc-PR is settled).

dougiesquire:
> I agree with @rabernat that serious update of docs is in order. But given that this will happen in a sister-PR soon then I'm also in favor of going forward with this.
> 
> (I'm happy to help with the docs-PR, but somebody needs to show me how to compile/view docs locally :))

I'm in the same boat here - help to help with #135 if/where I can, but don't know how best to contribute to the notebook.

roxyboy:
Seems like everyone is happy with this PR so I'm gonna merge this so that we can start working on the documentation :)

rabernat:
> Also note that there are breaking changes in this PR (e.g., `real`->`real_dim`) so a new minor release should follow (after doc-PR is settled).

Have we actually done a release that includes the `real` option? If so, we should do a deprecation cycle for this.

navidcy:
> > Also note that there are breaking changes in this PR (e.g., `real`->`real_dim`) so a new minor release should follow (after doc-PR is settled).
> 
> Have we actually done a release that includes the `real` option? If so, we should do a deprecation cycle for this.

Agreed. I think several releases lately included the `real` option in `dft`.

roxyboy:
> > > Also note that there are breaking changes in this PR (e.g., `real`->`real_dim`) so a new minor release should follow (after doc-PR is settled).
> > 
> > 
> > Have we actually done a release that includes the `real` option? If so, we should do a deprecation cycle for this.
> 
> Agreed. I think several releases lately included the `real` option in `dft`.

I've started PR #139 to address this.

PR #137, opened 2021-02-08
Refer to stable version of dependency docs

navidcy:
This PR updates the urls to the docs of for various dependency packages to point to the stable version rather than then latest one.


navidcy:
Btw, feel free to close this if you think it's not relevant. :)

rabernat:
I have not made up my mind whether this is the right choice.

Could you explain the reasoning behind pointing to stable over latest docs?

navidcy:
When I instal `xrft` would the most stable version of `dask` be installed as a dependency or the latest dev version? If the latter then pointing to the `docs/latest` is the way forward. If the former, then isn't it more appropriate to point to the docs of the version `xrft` will be using?

PR #139, opened 2021-02-17
Deprecation cycle for `real` flag

roxyboy:
I added `kwargs` to have a deprecation cycle for the `real` flag.


xrft/xrft.py
```diff
@@ -764,6 +780,10 @@ def cross_spectrum(
             + "with numpy-like behavior where the coordinates are disregarded."
         )
         warnings.warn(msg, FutureWarning)
+    elif "real" in kwargs:
```
rabernat, line 783: Why is this an `elif`. Couldn't we have both be true? Should probably just be a standalone `if` block.

PR #141, opened 2021-03-17
Energy correction for isotropic spectrum

roxyboy:
Added the flag to allow for energy/amplitude correction for isotropic spectrum as is done for the `power_spectrum()` and `cross_spectrum()`.

### DIFF
--- a/xrft/xrft.py
+++ b/xrft/xrft.py
@@ -1016,7 +1016,7 @@ def isotropic_power_spectrum(
     window_correction=False,
     window=None,
     nfactor=4,
-    **kwargs
+    **kwargs,
 ):
     """
     Calculates the isotropic spectrum from the
@@ -1115,7 +1115,7 @@ def isotropic_cross_spectrum(
     window_correction=False,
     window=None,
     nfactor=4,
-    **kwargs
+    **kwargs,
 ):
     """
     Calculates the isotropic spectrum from the
--- a/xrft/xrft.py
+++ b/xrft/xrft.py
@@ -1012,9 +1012,11 @@ def isotropic_power_spectrum(
     dim=None,
     shift=True,
     detrend=None,
-    density=True,
+    scaling="density",
+    window_correction=False,
     window=None,
     nfactor=4,
+    **kwargs
 ):
     """
     Calculates the isotropic spectrum from the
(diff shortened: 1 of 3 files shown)


xrft/xrft.py
```diff
@@ -1059,6 +1061,15 @@ def isotropic_power_spectrum(
...
     """
+    if "density" in kwargs:
+        density = kwargs.pop("density")
+        msg = (
+            "density flag will be deprecated in future version of xrft.isotropic_power_spectrum and replaced by scaling flag. "
+            + "density=True should be replaced by scaling='density' and "
+            + "density=False will not be maintained.\nscaling flag is ignored !"
+        )
+        warnings.warn(msg, FutureWarning)
+        scaling = "density" if density else "false_density"
```
rabernat, line 1072: Are the popping and duplicated messages necessary? Can't you just pass `density` and `window_correction` on to `power_spectrum` and have it take care of things?

xrft/xrft.py
```diff
@@ -1148,6 +1162,15 @@ def isotropic_cross_spectrum(
...
     """
+    if "density" in kwargs:
+        density = kwargs.pop("density")
+        msg = (
+            "density flag will be deprecated in future version of xrft.isotropic_cross_spectrum and replaced by scaling flag. "
+            + "density=True should be replaced by scaling='density' and "
+            + "density=False will not be maintained.\nscaling flag is ignored !"
+        )
+        warnings.warn(msg, FutureWarning)
+        scaling = "density" if density else "false_density"
```
rabernat, line 1173: Same

rabernat (commented):
I'm confused why it's necessary to do this in more than one place.

xrft/xrft.py
roxyboy: Makes sense. I'll remove the warning from the isotropic functions.

xrft/xrft.py
roxyboy: I think the popping should still be there so that `scaling` gets passed to `power_spectrum` even is `density` is called.
This allows for the `scaling` flag to be a default keyword in `isotropic_power_spectrum`

rabernat:
Thanks @roxyboy!

Issue #148, opened 2021-04-19
Erroneous behaviour when coordinates of transformation is in decreasing order


lanougue:
Hi, looking to #144 issue, here is a "bug" behaviour:
```
T = xr.tutorial.open_dataset("air_temperature").air.isel(time=[0, 1])
T_f1 = xrft.dft(T, dim="lat", true_phase=True, true_amplitude=True)
T = T.sortby('lat')
T_f2 = xrft.dft(T, dim="lat", true_phase=True, true_amplitude=True)
```

T_f1 and T_f2 should be the same but are not.

lanougue:
@roxyboy, @rabernat 
After a rapid check, as it is coded now, dft (and idft) can work correctly only if input data has increasing coordinates along the axis the dft is taken along. dft and idft need to be updated to take care of reversed coordinates.

From a performance point of view, I think that sorting can be slow (xr.DataArray.sortby) and only a "flip" is actually needed on the data since constant spacing is checked anyway. Actually fliping the data could probably be avoided by just creating a view before taking the actual fourier transform.

If we go like this, the _lag_coord() function has to be updated too.
What do you guys think ?
I can open a PR for that

roxyboy:
A PR would be great, thanks @lanougue ! :)

lanougue:
Done in #149 :)

lanougue:
Done in #149 :)

PR #149, opened 2021-04-20
Dft reverse axis

lanougue:
Handle decreasing coordinates in dft with true_phase = True.
Add a test to check


lanougue:
@roxyboy, merge ?

roxyboy:
I'm personally ok to merge but I think we should wait for @rabernat 's approval as well. He should be back from vacation next week.
We should also probably update the docs page using realistic data from the xarray tutorial (e.g. #148 ).

xrft/xrft.py
```diff
@@ -214,7 +214,10 @@ def _lag_coord(coord):
 
     v0 = coord.values[0]
     calendar = getattr(v0, "calendar", None)
-    lag = coord[(len(coord.data)) // 2]
+    reverse_coord = (
+        [] if (coord[-1] > coord[0]) else [-1]
+    )  # taking care of reverse coordinate (CAUTION : []  diff None)
```
rabernat, line 219: I don't understand this comment, specifically `(CAUTION : []  diff None)`.

xrft/xrft.py
```diff
@@ -214,7 +214,10 @@ def _lag_coord(coord):
...
+    lag = np.flip(coord.data, axis=reverse_coord)[(len(coord.data)) // 2]
```
rabernat, line 220: For the sake of the other devs, I would prefer to make this all a bit more explicit, i.e.
```python
if coord[-1] > coord[0]:
    coord_data = np.flip(coord.data, axis=-1)
else:
    coord_data = coord.data
lag = coord_data / (len(coord.data) // 2)
```

rabernat (changes requested):
Thanks so much for this PR @lanougue! I really appreciate your work.

I agree with @roxyboy that documentation of this behavior is needed in the public facing docstrings and / or narrative documentation.

xrft/xrft.py
lanougue: Sorry for this not being explicit. This is to emphasize that:
np.flip(a, axis=None) has a different behaviour than np.flip(a, axis=[]).
If axis=None, all axis are flipped.
If axis=[], no axis are flipped
I though it deserved to me mentioned in a comment

xrft/xrft.py
lanougue: ok but the last line should then be:
`lag = coord_data[(len(coord.data)) // 2]`

rabernat:
Thanks! So is the plan to defer the documentation updates to a future PR? If so, I will merge.

roxyboy:
> Thanks! So is the plan to defer the documentation updates to a future PR? If so, I will merge.

I give a thumbs up to a separate PR for docs :)

Issue #147, opened 2021-04-19
Erroneous examples in documentation


lanougue:
Hello everybody,
I wasn't aware of the new realase ! I just had a look. Thanks all for the excellent job!
I gave a look in the doc at "example of discrete and inverse discrete Fourier transform" and recognized my favorite example ;) However I found something a bit erroneous. In cell [13] of the notebook, the numpy fft (npft) has to be 1) "fftshifted", 2) multiplied by the "r" factor and 3) multiplied by "dx" before plotting in order to look-like the theoretical transform.
I can understand about the "dx" for the amplitude, however the "r" factor (and the fftshift)  is a pure fancyful factor I had some hard time to remove in dft. I think this example totaly minimize the phase impact in fft and mislead the reader since what is plotted is not really the numpy (or xrft) fft output.

To clarify as much as possible the documentation, I think this example should exactly point the difference between xrft.fft, xrft.dft and numpy.fft

I think the forward (dft and fft) and backward transformation (idft and ifft) desesrves a decicated different example documentation.

I can help for this if needed

rabernat:
Hi @lanougue! Thanks a lot for checking this.

We would really appreciate your contributions to the documentation. Since you led the implementation of the "shift" features, it makes sense for you to be the one to update the documentation here. Feel free to propose whatever changes you think are needed via PR.

lanougue:
Hi @rabernat. Yes, I will work on it. 

In the future,iIt would be nice if I could receive an email/notification when someone post an issue in xrft. Is there any list to subscribe to ?

roxyboy:
PR #151 addressed this so I'm going to close this :)

Issue #144, opened 2021-03-21
Doubt about behavior of `dft`


tomchor:
Great thing that you guys have added inverse `fft`s! However, I can't figure out how to exactly preserve the original coordinates of the data though. I was able to do that with `fft`s so I'm guessing there should be also a way to do this with `xrft`. Here's a MWE:

```python
In [105]: import xarray as xr

In [106]: import xrft

In [107]: T = xr.tutorial.open_dataset("air_temperature").air.isel(time=[0, 1])

In [108]: T_f = xrft.dft(T, dim="lat", true_phase=True, true_amplitude=True)

In [109]: T2 = xrft.idft(T_f, dim="freq_lat", true_phase=True, true_amplitude=True,)

In [110]: T.lat
Out[110]: 
<xarray.DataArray 'lat' (lat: 25)>
array([75. , 72.5, 70. , 67.5, 65. , 62.5, 60. , 57.5, 55. , 52.5, 50. , 47.5,
       45. , 42.5, 40. , 37.5, 35. , 32.5, 30. , 27.5, 25. , 22.5, 20. , 17.5,
       15. ], dtype=float32)
Coordinates:
  * lat      (lat) float32 75.0 72.5 70.0 67.5 65.0 ... 25.0 22.5 20.0 17.5 15.0
Attributes:
    standard_name:  latitude
    long_name:      Latitude
    units:          degrees_north
    axis:           Y

In [111]: T2.lat
Out[111]: 
<xarray.DataArray 'lat' (lat: 25)>
array([-30. , -27.5, -25. , -22.5, -20. , -17.5, -15. , -12.5, -10. ,  -7.5,
        -5. ,  -2.5,   0. ,   2.5,   5. ,   7.5,  10. ,  12.5,  15. ,  17.5,
        20. ,  22.5,  25. ,  27.5,  30. ])
Coordinates:
  * lat      (lat) float64 -30.0 -27.5 -25.0 -22.5 -20.0 ... 22.5 25.0 27.5 30.0
Attributes:
    spacing:  2.4999999999999964
```

Using the `xrft.fft` option also doesn't produce what I expected:

```python
In [112]: T_f = xrft.fft(T, dim="lat",)

In [113]: T2 = xrft.ifft(T_f, dim="freq_lat")

In [114]: T.lat
Out[114]: 
<xarray.DataArray 'lat' (lat: 25)>
[...]

PR #151, opened 2021-05-20
Updates to documentation with xarray tutorial data

roxyboy:
A PR to update the documentation and address #147 . I added an example from #144.


roxyboy:
I think we should do another release if people are ok with the new documentation.

roxyboy:
@lanougue @rabernat Any thoughts on this PR on making the documentation more helpful? I added an example using the `xarray` tutorial data.

doc/DFT-iDFT_example.ipynb
```diff
@@ -9,7 +9,7 @@
   },
```
lanougue, line 9: <p>I think that it should be mentioned that numpy fft has been fftshifted in the plot</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/xgcm/xrft/pull/151/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

lanougue, line 9: <p>I think it should be mentioned that xrft.fft and numpy.fft lack the good amplitudes! They both have to me multiplied by dx to match the theory.</p><p>xrft.dft automatically do this with the flag true_amplitude=True</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/xgcm/xrft/pull/151/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

lanougue, line 9: <p>I think we should had the same coments as before. ifft misses the amplitude scaling (dk or df) but since it is only the inverse of (N_point*dx), the plot looks right.</p><p>The proof is that first plot is (da_fft*dx) and in this plot: this is ifft(da_fft) which is shown.</p><p> </p><p> </p><p>idft automatically do the right scaling (true_amplitude) whitout needing to take care of the number of points.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/xgcm/xrft/pull/151/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

lanougue, line 9: <p>I would replace "i.e. the imaginary part" by "the complex term"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/xgcm/xrft/pull/151/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

lanougue, line 9: <p>Your comment is clear to me. However, instead of plotting x[nshift:], I would plot what any user would plot ie: (x,inda_fft). This would emphasizes that ifft actually do not care about the phase/shifting and do not recover the right position of the signal.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/xgcm/xrft/pull/151/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

lanougue:
Thanks @roxyboy for this work! A personnal global comment on this documentation would be that it seems to prove that numpy.fft could be enough if used carrefully (in combination with apply_ufunc for example).
On the contrary I would promote xrft power arguing that it takes care of phase, amplitude, spacing, number of points, ... automatically. From my point of view, this should also be done by emphasizing the defaults of the numpy.fft instead of trying to show that they can be handled by a careful user.
xrft is not only a very useful wrapper for Fourier Transforms with xarrays, but also correctly takes care of all the scaling and phasing inherent to Fourier Transforms.
Thanks again for the energy you put in xrft! :)

doc/DFT-iDFT_example.ipynb
roxyboy: <p>Not sure I understand what you're meaning by <code>(N_point*dx)</code>...</p>

roxyboy: Can you take a look at the new commit and see if you're ok with my edits along with your point above?

lanougue: fft misses the dx scaling and ifft misses the df scaling. When multiplying the two errors by doing ifft(fft)) ie dx*df=1/N, there is a 1/N missing wich is, in fact, arbitrarily included in the ifft definition.
If you check the mathematical definitions : fft and ifft defer not only in the sign in the exponentials but also in the scaling. There is a 1/N_point in ifft

lanougue: By doing the righ scalings in dft and idft, there is no more consideration of the number of points.

roxyboy: Ok, makes sense and added it to the documentation.

lanougue, line 9: <p>I just realize that I forgot the 2 pi factor which is included in the fft/ifft definitions ! Thus, fft misses the dx scaling and ifft misses the df (and not dk) definition (dk = 2 pi df). When applying ifft(fft)), the 2 pi is included in ifft definition (which corresponds to the continuous version on the inverse transform (see my other comment)). And the 1/N is, as you explained, the normalization factor.</p><p>Maybe, you can link to this page : https://numpy.org/doc/stable/reference/routines.fft.html?highlight=fft#module-numpy.fft</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/xgcm/xrft/pull/151/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

roxyboy: The link is already there.

lanougue: <p>Ok about the link... sorry</p><p> </p><p>The "i" are missing in the exponentials of all expression.</p><p>Maybe replacing "k" by "f". This is the usual convention when there is "2 pi" in the exponentials of the Fourier Transforms expressions.</p>

roxyboy: Done :)

lanougue: <p>two missing k-&gt;f here too</p>

lanougue, line 9: <p>There is two missing k -&gt; f</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/xgcm/xrft/pull/151/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

Issue #145, opened 2021-04-01
Isotropize returns unexpected range of wavenumbers


wenegrat:
Thanks for working on this project, it seems like it will be very useful. 

I wanted to flag what seems to me to be unexpected behavior from isotropize.

Specifically the calculation of a radial wavenumber as, 

>     freq_r = np.sqrt(k ** 2 + l ** 2).rename("freq_r")

gives a vector where the highest wavenumber exceeds the Nyquist wavenumber. For example if the sample spacing is 1,  this gives a maximum wavenumber of 0.7. This is because the isotropic spectrum is being calculated out into the corner of the Cartesian grid, for instance outside the black line in the following plot:

![download-2](https://user-images.githubusercontent.com/6423335/113357216-9ea0d680-9311-11eb-8185-85decab3aee1.png)

This to me seems like undesirable behavior, and for instance it can result in a rapid roll of energy above the k, l Nyquist frequency (as higher than this you only have energy from wavenumbers oriented towards the corners of the Cartesian grid.

I would think the expected behavior would follow that from calc_ispec in @cesar-rocha's package [pyspec](https://github.com/pyspec/pyspec), which I reproduce below to show the difference.

> def calc_ispec(k,l,E,ndim=2):
    """ Calculates the azimuthally-averaged spectrum
        Parameters
        ===========
        - E is the two-dimensional spectrum
        - k is the wavenumber is the x-direction
        - l is the wavenumber in the y-direction
        Output
        ==========
        - kr: the radial wavenumber
        - Er: the azimuthally-averaged spectrum """

    dk = np.abs(k[2]-k[1])
    dl = np.abs(l[2]-l[1])

    k, l = np.meshgrid(k,l)

    wv = np.sqrt(k**2+l**2)

    if k.max()>l.max():
        kmax = l.max()
    else:
        kmax = k.max()

    if ndim==3:
        nl, nk, nomg = E.shape
    elif ndim==2:
        nomg = 1

    dkr = np.sqrt(dk**2 + dl**2)
    kr =  np.arange(dkr/2.,kmax+dkr/2.,dkr)
    Er = np.zeros((kr.size,nomg))

    for i in range(kr.size):

        fkr =  (wv>=kr[i]-dkr/2) & (wv<=kr[i]+dkr/2)
        dth = np.pi / (fkr.sum()-1)
        if ndim==2:
            Er[i] = (E[fkr]*(wv[fkr]*dth)).sum()
        elif ndim==3:
[...]

PR #152, opened 2021-05-31
Truncate isotropic spectra beyond Nyquist wavenumber

roxyboy:
This addresses issue #145 by truncating the isotropic spectra for wavenumbers higher than the Nyquist wavenumber.


roxyboy:
The truncation is implemented after the isotropic binning is done: https://github.com/roxyboy/xrft/blob/93238b5725eb1ac3effa2c946b281681d6f2d77e/xrft/xrft.py#L1004

roxyboy:
@rabernat Could I ask you to review this?

xrft/xrft.py
```diff
@@ -971,7 +972,7 @@ def _groupby_bins_agg(
     return result
 
 
-def isotropize(ps, fftdim, nfactor=4):
+def isotropize(ps, fftdim, nfactor=4, **kwargs):
```
rabernat, line 975: Curious why you have added the generic `**kwargs` rather than `truncate=None` here? Remember [explicit is better than implicit](https://www.python.org/dev/peps/pep-0020/#id2).

xrft/xrft.py
```diff
@@ -1098,7 +1123,7 @@ def isotropic_power_spectrum(
 
     fftdim = ["freq_" + d for d in dim]
 
-    return isotropize(ps, fftdim, nfactor=nfactor)
+    return isotropize(ps, fftdim, nfactor=nfactor, kwargs={"truncate": truncate})
```
rabernat, line 1126:
```suggestion
    return isotropize(ps, fftdim, nfactor=nfactor, truncate=truncate)
```

xrft/xrft.py
```diff
@@ -1197,7 +1226,7 @@ def isotropic_cross_spectrum(
 
     fftdim = ["freq_" + d for d in dim]
 
-    return isotropize(cs, fftdim, nfactor=nfactor)
+    return isotropize(cs, fftdim, nfactor=nfactor, kwargs={"truncate": truncate})
```
rabernat, line 1229:
```suggestion
    return isotropize(cs, fftdim, nfactor=nfactor, truncate=truncate)
```

xrft/tests/test_xrft.py
```diff
@@ -928,7 +929,9 @@ def test_isotropize(N=512):
     def _test_iso(theta):
         ps = xrft.power_spectrum(theta, spacing_tol=spacing_tol, dim=dims)
         ps = np.sqrt(ps.freq_x ** 2 + ps.freq_y ** 2)
-        ps_iso = xrft.isotropize(ps, fftdim, nfactor=nfactor)
+        ps_iso = xrft.isotropize(
+            ps, fftdim, nfactor=nfactor, kwargs={"truncate": truncate}
```
rabernat, line 933:
```suggestion
            ps, fftdim, nfactor=nfactor, truncate=truncate
```

rabernat (commented):
Thanks for doing this Takaya. I'm not sure I understand why you are using `kwargs` here though.

xrft/xrft.py
roxyboy: The idea I had was the we'd eventually drop the truncate flag with the behavior always being that we truncate for the wavenumber. I kept it in `kwargs` with this in mind as an intermediate step until we change the default behavior but I'll move it to an explicit input.

rabernat: I see. That makes sense. But I think we can handle that just by setting `truncate=True` in a future release.

roxyboy:
Can we merge this @rabernat ??

PR #155, opened 2021-06-15
A deprecation cycle to unify fft and dft by dropping the latter

roxyboy:
A PR to avoid the confusion between `fft` and `dft` by adding a deprecation warning to when `dft` is called.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,22 +4,22 @@ xrft: Fourier transforms for xarray data
 |pypi| |conda forge| |conda-forge| |Build Status| |codecov| |docs| |DOI| |license| |Code style|
 
 **xrft** is an open-source Python package for
-taking the discrete Fourier transform (DFT) on xarray_ and dask_ arrays.
+taking the Fast Fourier transform (FFT) on xarray_ and dask_ arrays.
 
 .. _xarray: http://xarray.pydata.org/en/stable/
 .. _dask: https://dask.org
 
 It is:
 
-- **Powerful**: It keeps the metadata and coordinates of the original xarray dataset and provides a clean work flow of DFT.
+- **Powerful**: It keeps the metadata and coordinates of the original xarray dataset and provides a clean work flow of FFT.
 - **Easy-to-use**: It uses the native arguments of `numpy FFT`_ and provides a simple, high-level API.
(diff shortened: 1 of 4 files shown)


lanougue:
Hi @roxyboy and thanks for doing this PR.

I now remember one of our old conversation when you mentioned that xrft.fft should always (because of its name) mimic numpy.fft behaviour. The direction xrft.fft is now going to, make me think that it will not be the case anymore.
However, I fully support this new direction since I beleive that xrft.fft will be much more complient with the Fourier Transformation definition.
We may nevertheless think about keeping or not fft as the main name

roxyboy:
> I now remember one of our old conversation when you mentioned that xrft.fft should always (because of its name) mimic numpy.fft behaviour. The direction xrft.fft is now going to, make me think that it will not be the case anymore.
> However, I fully support this new direction since I beleive that xrft.fft will be much more complient with the Fourier Transformation definition.
> We may nevertheless think about keeping or not fft as the main name

Having a deprecation cycle as I do in this PR should alleviate the stress in dropping `dft`. People who come to `xrft` are likely looking for something that augments `numpy.fft` so I think new users will realize that `xrft.fft` and `numpy.fft` are not the same.

lanougue:
In current official version of xrft, when calling dft, you get sometimes the warning saying:  "Consider using xrft.fft to ensure future compatibility with numpy.fft like behavior and to deactivate this warning" ...

But I agree with you and believe this is a good time to augment fft capabilities with xarray/xrft (and drop dft) even if it kind of  breaks some usual behaviour. A good documentation should help the transition

roxyboy:
> In current official version of xrft, when calling dft, you get sometimes the warning saying: "Consider using xrft.fft to ensure future compatibility with numpy.fft like behavior and to deactivate this warning" ...
> 
> But I agree with you and believe this is a good time to augment fft capabilities with xarray/xrft (and drop dft) even if it kind of breaks some usual behaviour. Deprecation cycle and a good documentation should help the transition.
> 
> In any case numpy.fft like behaviour can easily be recovered with true_phase/amplitude set to False

I think the PR won't break any previous behavior... Only the warning messages will change.

roxyboy:
All the new PR merges has resulted in a conflict so I'm gonna do this in a new PR.

PR #157, opened 2021-06-15
Storing lag for round-tripping

lanougue:
In this PR, I store the lag of the transformed dimension as an attribute of the new coordinates (same as "spacing").
This seems logical (to me) and ensures that this attributes will not be accidentally erased when applying some transformation/operation (filtering most probably) on the transformed array.


lanougue:
@roxyboy, I am really not smart enough to understand these pre-commit fails...
Any suggestion ?

xrft/xrft.py
```diff
@@ -539,8 +540,9 @@ def idft(
             dim = [d for d in dim if d != real_dim] + [
                 real_dim
             ]  # real dim has to be moved or added at the end !
-
-    if lag is not None:
+    if lag is None:
```
roxyboy, line 544: This is a great start. Thanks for the PR. I think we want to go one step further and now use this "direct_lag" object in xrft.idft to get out the original coordinates in `da`, i.e. da.coords = xrft.idft(xrft.dft(da)).coords.

roxyboy (changes requested):

roxyboy:
> @roxyboy, I am really not smart enough to understand these pre-commit fails...
> Any suggestion ?

Probably just has to do with black formatting. Running `black xrft.py` should fix it (assuming you have black installed).

xrft/xrft.py
lanougue: Yes, this is exactly what this PR is doing.
Now DA =  xrft.idft(xrft.dft(da)) will have the same coordinates as da
(when using true_phase = true in both calls)

lanougue:
> > @roxyboy, I am really not smart enough to understand these pre-commit fails...
> > Any suggestion ?
> 
> Probably just has to do with black formatting. Running `black xrft.py` should fix it (assuming you have black installed).

When I commit, black says that it passes the test (when black test fails, black corrects the files and the test passes when doing a new commit). I thus do not understand why it fails on github ...

xrft/xrft.py
roxyboy: Great! Can you add a test to check this??

lanougue: Sure ! Done

xrft/tests/test_xrft.py
```diff
@@ -1261,7 +1261,7 @@ def test_idft_dft():
     )
     xrt.assert_allclose(s, IFTs)
 
-    ## when lag is set automattically
+    ## when lag is set automatically
     IFTs = xrft.idft(FTs, shift=True, true_phase=True, true_amplitude=True)
     xrt.assert_allclose(s, IFTs)
```
roxyboy, line 1266: Sorry, this is an elementary question but does xarray.testing also check for the coordinates?

lanougue: Yes, I verified manually.

However, with these changes default behaviour of idft has changed. By default (lag=None), output coordinates were centered. Now, the default is to use the direct_lag value. Maybe it deserves some warning

roxyboy: Mmm, I feel like the `idft` functionality hasn't been around long enough to warrant a deprecation cycle so I think a warning would suffice. @rabernat Would you agree with this?

roxyboy:
@lanougue Is this ready to merge?

lanougue:
yes !

Issue #159, opened 2021-06-30
`Detrend='linear'` with `chunks_for_segments=True` fails


roxyboy:
I haven't gotten to the bottom of this yet by the combination of `Detrend='linear'` and `chunks_for_segments=True` fails. Below is a reproducible example:

```
N = 32
da = xr.DataArray(
        np.random.rand(N, N, N),
        dims=["time", "y", "x"],
        coords={"time": range(N), "y": range(N), "x": range(N)},
    )
xrft.dft(
        da.chunk({"y":16,"x":16}).fillna(0.), dim=["y",'x'], detrend='linear', chunks_to_segments=True
    )
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-82-d97abd488e86> in <module>
      5         coords={"time": range(N), "y": range(N), "x": range(N)},
      6     )
----> 7 xrft.dft(
      8         da.chunk({"y":16,"x":16}).fillna(0.), dim=["y",'x'], detrend='linear', chunks_to_segments=True
      9     )

/srv/conda/envs/notebook/lib/python3.8/site-packages/xrft/xrft.py in dft(da, spacing_tol, dim, real_dim, shift, detrend, window, true_phase, true_amplitude, chunks_to_segments, prefix, **kwargs)
    411         f = fft_fn(fft.ifftshift(da.data, axes=axis_num), axes=axis_num)
    412     else:
--> 413         f = fft_fn(da.data, axes=axis_num)
    414 
    415     if shift:

/srv/conda/envs/notebook/lib/python3.8/site-packages/dask/array/fft.py in func(a, s, axes)
    180         for each_axis in axes:
    181             if len(a.chunks[each_axis]) != 1:
--> 182                 raise ValueError(chunk_error % (each_axis, a.chunks[each_axis]))
    183 
    184         chunks = out_chunk_fn(a, s, axes)

ValueError: Dask array only supports taking an FFT along an axis that 
has a single chunk. An FFT operation was tried on axis 2 
which has chunks (1, 1). To change the array's chunks use dask.Array.rechunk.
```

It seems that the linear detrending is affecting the chunk size. The snippet above works fine when `detrend='None'` or `'constant'`.

roxyboy:
This is originating from upon applying linear detrending, the ordering of the dimension changes so the `axis_num` becomes inconsistent with the actual dimensions the FFT is going be take over.

roxyboy:
PR #160 fixed this.

PR #160, opened 2021-06-30
Update `axis_num` when `detrend='linear'` is taken

roxyboy:
This addresses issue #159 and updates the `axis_num` when `detrend='linear'` because the latter changes the order of the dimensions.

### DIFF
--- a/xrft/xrft.py
+++ b/xrft/xrft.py
@@ -411,8 +411,10 @@ def dft(
         delta_x.append(delta)
         lag_x.append(lag)
 
-    if detrend:
+    if detrend is not None:
         da = _detrend(da, dim, detrend_type=detrend)
+        # Update the axis_num upon detrending
+        axis_num = [da.get_axis_num(d) for d in dim]
 
     if window is not None:
         _, da = _apply_window(da, dim, window_type=window)
--- a/xrft/xrft.py
+++ b/xrft/xrft.py
@@ -413,7 +413,8 @@ def dft(
 
     if detrend is not None:
         if detrend == "linear":
-            da = _detrend(da, dim, detrend_type=detrend).transpose(*rawdims)
+            orig_dims = da.dims
+            da = _detrend(da, dim, detrend_type=detrend).transpose(*orig_dims)
         else:
             da = _detrend(da, dim, detrend_type=detrend)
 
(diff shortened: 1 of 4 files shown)


lanougue:
Hi @roxyboy. Instead of updating axis_num after detrending, I would rather transpose the output of detrending into the previous/old axis number. This would ensure compatibility with potential other variables defined previously with the initial order (like the "N" variable used in _freq() function after detrending).
As it is now, I think it can create errors

xrft/xrft.py
```diff
@@ -412,9 +412,10 @@ def dft(
         lag_x.append(lag)
 
     if detrend is not None:
-        da = _detrend(da, dim, detrend_type=detrend)
-        # Update the axis_num upon detrending
-        axis_num = [da.get_axis_num(d) for d in dim]
+        if detrend == 'linear':
```
roxyboy, line 415: @lanougue I implemented your suggestion.

lanougue (changes requested):
Not sure about this because da's dims order is altered when real_dim is not None (see some code lines before your changes).
I would save da's dims order just before detrending and transpose just after it.

```
dims_order = da.dims
da = _detrend(da, ...).transpose(*dims_order)
```

roxyboy:
> Not sure about this because da's dims order is altered when real_dim is not None (see some code lines before your changes).
> I would save da's dims order just before detrending and transpose just after it.
> 
> ```
> dims_order = da.dims
> da = _detrend(da, ...).transpose(*dims_order)
> ```

Yea, that's why I've changed the order to save `rawdims` after the `real_dim` argument. `rawdims` and your `dims_order` should be the same with this PR.

xrft/xrft.py
```diff
@@ -557,13 +558,13 @@ def idft(
...
         daft = _stack_chunks(daft, dim)
 
-    rawdims = daft.dims  # take care of segmented dimesions, if any
-
     if real_dim is not None:
         daft = daft.transpose(
             *[d for d in daft.dims if d not in [real_dim]] + [real_dim]
         )  # dimension for real transformed is moved at the end
 
+    rawdims = daft.dims  # take care of segmented dimensions, if any
```
roxyboy, line 566: @lanougue This is where I've changed the order when `rawdims` is saved.

lanougue:
raw_dims definition (as it was before this change) ensures input and output have the same dims order. If you move raw_dims as you did, this behaviour will be broken because the real dim will always be moved to the end

xrft/xrft.py
lanougue: moving real_dim to the end of the list is a necessity because of rfft function. However, I beleive it should not affect user input dims order (which was ensured by saving raw_dims). This change will break this behaviour.

roxyboy:
> raw_dims definition (as it was before this change) ensures input and output have the same dims order. If you move raw_dims as you did, this behaviour will be broken because the real dim will always be moved to the end

Yea, you're correct. I added an `orig_dims` variable before detrending.

roxyboy:
So... I guess I'll merge this.

PR #161, opened 2021-07-02
Deprecating `dft` and replacing it with `fft`

roxyboy:
A PR to initiate a deprecation cycle for `xrft.dft` and to replace it with `xrft.fft`. As the tests all pass using `dft`, none of the previous behaviors should break with only the warning messages changing.

### DIFF
--- a/xrft/xrft.py
+++ b/xrft/xrft.py
@@ -418,7 +418,8 @@ def fft(
             da.get_axis_num(d) for d in dim if da[d][-1] < da[d][0]
         ]  # handling decreasing coordinates
         f = fft_fn(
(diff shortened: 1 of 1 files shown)


roxyboy:
@santisoler Do you have any input on this. I've tried to follow your suggestion of unifying `xrft` to `fft`.

santisoler (approved):
Looks good to me!

roxyboy:
@rabernat This deprecation cycle of removing `dft` and unifying the default function to `fft` will result in quite a drastic change down the line but would you agree with this? It seems that having two functions (`fft` and `dft`) is confusing to users.

Issue #162, opened 2021-08-24
Possible error in handling of coords when using `real_dim`


paigem:
I ran into an error when running xrft functions. The error seems to occur when the xarray DataArray has multiple coordinates that depend on the dimension that is passed to the `real_dim` flag. Below is (my attempt at) a minimal reproducible example that yields an error included just below.

```python
# Create dummy data array
data = np.arange(-100,100) 

# Add coordinate time and time2, both of which are attached to dimension 'time'
coord = {'time':np.linspace(0,10,len(data)),'time2':('time',np.arange(len(data)))}

# Define the DataArray
da = xr.DataArray(data,dims='time',coords=coord)

# Compute dft
xrft.dft(da,dim='time',real_dim='time')
```

<details><summary>Value Error</summary>

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-60-ee5ce90d1fc2> in <module>
----> 1 xrft.dft(da,dim='time',real_dim='time')

/srv/conda/envs/notebook/lib/python3.8/site-packages/xrft/xrft.py in dft(da, spacing_tol, dim, real_dim, shift, detrend, window, true_phase, true_amplitude, chunks_to_segments, prefix, **kwargs)
    438 
    439     newcoords, swap_dims = _new_dims_and_coords(da, dim, k, prefix)
--> 440     daft = xr.DataArray(
    441         f, dims=da.dims, coords=dict([c for c in da.coords.items() if c[0] not in dim])
    442     )

/srv/conda/envs/notebook/lib/python3.8/site-packages/xarray/core/dataarray.py in __init__(self, data, coords, dims, name, attrs, indexes, fastpath)
    408             data = _check_data_shape(data, coords, dims)
    409             data = as_compatible_data(data)
--> 410             coords, dims = _infer_coords_and_dims(data.shape, coords, dims)
    411             variable = Variable(dims, data, attrs, fastpath=True)
    412             indexes = dict(

/srv/conda/envs/notebook/lib/python3.8/site-packages/xarray/core/dataarray.py in _infer_coords_and_dims(shape, coords, dims)
    156         for d, s in zip(v.dims, v.shape):
    157             if s != sizes[d]:
--> 158                 raise ValueError(
    159                     f"conflicting sizes for dimension {d!r}: "
    160                     f"length {sizes[d]} on the data but length {s} on "

ValueError: conflicting sizes for dimension 'time': length 101 on the data but length 200 on coordinate 'time2'
[...]

PR #163, opened 2021-08-26
Raise error for non-dimensional coordinates

paigem:
This PR has two additions (see issue #162):

- raises an error in `xrft.fft()`  when there are multiple coordinates attached to the same dimensions (added in `xrft.py`)
- tests for the above error (added to `test_xrft.py`)

Note that I haven't contributed code for awhile, so a thorough read-through of what I've done would be great. Also, I have never used 'code-style' before, but I got a message stating that "all jobs have failed" and I'm not quite sure how to address this. So let me know if I need to update something here!

### DIFF
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -351,8 +351,6 @@ def test_window_single_dim():
     ps.load()
 
 
-
-
 class TestSpectrum(object):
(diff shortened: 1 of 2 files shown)


roxyboy:
Thanks @paigem, I'll go ahead and merge this.

PR #168, opened 2021-11-25
Docs

zmoon:
Update docs, configure some things that were not yet configured, ...
* clean up Conda env
* fix GH issue links in whats-new
* RTD theme set in conf
* add intersphinx/crossreferences
* clean up docstrings
* main API page organized; use top-level namespace for the generated documentation of the functions
* edit other text a bit

Remaining:
* [ ] example nbs


xrft/xrft.py
```diff
@@ -887,10 +902,10 @@ def cross_spectrum(
 
 
 def cross_phase(da1, da2, dim=None, true_phase=False, **kwargs):
-    """
-    Calculates the cross-phase between da1 and da2.
+    r"""
```
roxyboy, line 946: Is 'r' necessary here?

roxyboy:
Thanks for these edits @zmoon ! :)

xrft/xrft.py
```diff
@@ -271,6 +271,7 @@ def idft(
 
 def fft(
     da,
+    *,
```
roxyboy, line 274: What's this '*' for?

roxyboy:
@zmoon Can you run black on `detrend.py`? The pre-commit tests seems to be failing...

xrft/xrft.py
zmoon: To enforce the arguments after it being used as keyword arguments, i.e., only `da` is allowed to be passed as positionally. This seems to already be the convention in the examples and such, adding the `*` just enforces it. If you are interested, though, I could also add it to the other functions. Or can remove.

xrft/xrft.py
zmoon: It allows the math to be written without extra backslash-escaping, that's why I added it.

image
:point_up: original, with the `\t` in `\text` lost due to be converted to tab

zmoon:
@roxyboy, when are you planning to make the `true_*` arguments `True` by default? Just wondering, as currently the examples generate warnings related to that.

Also would you mind if I switch from `nbsphinx` to `myst_nb` to render the nbs? `nbsphinx` is not doing well in the cases with inline code inside strong or links, e.g. in [the main example](https://xrft.readthedocs.io/en/latest/DFT-iDFT_example.html). And with `myst_nb`, we can easily cross-reference API and use admonitions etc.

xrft/xrft.py
```diff
@@ -552,7 +580,12 @@ def ifft(
...
     if lag is None:
         lag = [daft[d].attrs.get("direct_lag", 0.0) for d in dim]
-        msg = "Default idft's behaviour (lag=None) changed! Default value of lag was zero (centered output coordinates) and is now set to transformed coordinate's attribute: 'direct_lag'."
+        msg = (
+            "Default idft's behaviour (lag=None) changed! "
+            "Default value of lag was zero (centered output coordinates) "
+            "and is now set to transformed coordinate's attribute 'direct_lag', "
+            "defaulting to zero if that attribute is not set."
+        )
         warnings.warn(msg, FutureWarning)
```
zmoon, line 589: Should this still be `FutureWarning` given that it has already been changed?

zmoon:
Also how big is the dataset for the MITgcm example? Could it be included in the repo or somewhere else where it could be downloaded?

xrft/xrft.py
roxyboy: The defaults is still `False` for these flags I think.

roxyboy:
> @roxyboy, when are you planning to make the `true_*` arguments `True` by default? Just wondering, as currently the examples generate warnings related to that.

This is a good question. What's the normal period packages allow for a deprecation cycle...?

xrft/xrft.py
zmoon: The default `lag` is still `None`, but the behavior has changed slightly (https://github.com/xgcm/xrft/commit/ca652c983ab593297377ffcf96d81e7ccf7429cd), since it now tries the `default_lag` attr before defaulting to zero. So maybe just normal `UserWarning` instead of `FutureWarning`?

roxyboy: Ok sure, I'm ok with `UserWarning`

KevinYanesG:
Grat work! I've found some small typos in the notebooks of `[zmoon:docs](https://github.com/zmoon/xrft/tree/docs)`. I cannot comment it directly here, because the diff is too large. What is the best-practice-procedure here? Should I fork [zmoon:docs](https://github.com/zmoon/xrft/tree/docs) and pull a request into his repo?

KevinYanesG:
I wanted to correct the other two notebooks that haven't been refactored in this PR: https://github.com/xgcm/xrft/pull/185.

DWesl:
> > @roxyboy, when are you planning to make the `true_*` arguments `True` by default? Just wondering, as currently the examples generate warnings related to that.
> 
> This is a good question. What's the normal period/cycles packages allow for a deprecation...?

- [NumPy](https://numpy.org/neps/nep-0023-backwards-compatibility.html#implementing-deprecations-and-removals): The longer of two releases or one year
- [SciPy](https://docs.scipy.org/doc/scipy/dev/core-dev/index.html#deprecations): Six months, usually two releases
- [Python](https://peps.python.org/pep-0387/#making-incompatible-changes): At least two minor versions (~2 years), ideally five years

PR #171, opened 2021-11-29
Fix `requires` in `setup.cfg`

zmoon:
* removed `setuptools` from `install_requires`
* `setup_requires` info is specified in `pyproject.toml`, so removed it
* added initial groups for `cftime` and `numpy_groupies`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -14,6 +14,7 @@ classifiers =
     License :: OSI Approved :: MIT License
     Natural Language :: English
     Operating System :: OS Independent
+    Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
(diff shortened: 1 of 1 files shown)


rabernat:
Thanks a lot for this PR @zmoon! LGTM!

setup.cfg
```diff
@@ -34,26 +35,25 @@ install_requires =
...
 
 [aliases]
-test=pytest
+test = pytest
 
 [options.extras_require]
+io =
+    cftime
+agg =
+    numpy_groupies
```
rabernat, line 50: Would be nice to document these options in the documentation somewhere.

zmoon: Yeah, and maybe `iso` would be a better for the `agg` one since it is used for the isotropic functions. Also currently `numpy_groupies` is required for `isotropize`, so maybe it should be a required dep, or instead of a fallback to xarray groupby-bins should be implemented. Or wait, since it seems like xarray may support numpy-groupies for groupby in the near future.

roxyboy: Indeed, we will only depend on `numpy_groupies` until it is incorporated in `xarray` upstream.

roxyboy:
So, I guess I'll merge this along with #168 ?

zmoon:
I still have a bit more I want to do in #168. And there is still the question of the MITgcm data file. But I guess that could be separate.

roxyboy:
> I still have a bit more I want to do in #168. And there is still the question of the MITgcm data file. But I guess that could be separate.

Ok, in that case, I'll merge just #171

PR #174, opened 2022-01-13
Readded test_xrft.py

avik2007:
I fixed a typo in cross_spectrum that made using window_correction = True unusable with the function. I added 
a section to test_cross_spectrum that should test that this fix works.

### DIFF
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -525,7 +525,7 @@ class TestSpectrum(object):
         cs = xrft.cross_spectrum(
             da, da2, dim=dim, shift=True, window="hann", detrend="constant"
         )
-        test = (daft * np.conj(daft2)).values / N ** 4
+        test = (daft * np.conj(daft2)) / N ** 4
 
         dk = np.diff(np.fft.fftfreq(N, 1.0))[0]
         test /= dk ** 2
@@ -541,10 +541,11 @@ class TestSpectrum(object):
             detrend="constant",
             window_correction=True,
         )
-        test = (daft * np.conj(daft2)).values / N ** 4
+        test = (daft * np.conj(daft2)) / N ** 4
         window, _ = _apply_window(da, dim, window_type="hann")
         dk = np.diff(np.fft.fftfreq(N, 1.0))[0]
-        test /= dk ** 2 * (window ** 2).mean()
+        test /= dk ** 2
+        test /= (window ** 2).mean()
 
         npt.assert_almost_equal(cs.values, test)
         npt.assert_almost_equal(np.ma.masked_invalid(cs).mask.sum(), 0.0)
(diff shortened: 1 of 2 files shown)


xrft/tests/test_xrft.py
```diff
@@ -541,7 +541,7 @@ def test_cross_spectrum(self, dask):
             window_correction=True,
         )
         test = (daft * np.conj(daft2)).values / N ** 4
-        window, _ = _apply_window(da, dim, window_type="hann")
+        window, _ = xrft._apply_window(da, dim, window_type="hann")
```
roxyboy, line 544: The underscore `_` in front of the function indicates that it's an implicit function and this is probably the cause for the error you're getting. Can you try importing it at the top of the `test_xrft.py` file: `from ..xrft import _apply_window`?

roxyboy: You're gonna need to remove the `xrft` at the head of `_apply_window` since this is what you imported.

avik2007: @roxyboy I added the fix. let me know if there's anything else I have forgotten.

xrft/tests/test_xrft.py
```diff
@@ -542,7 +542,7 @@ def test_cross_spectrum(self, dask):
             window_correction=True,
         )
         test = (daft * np.conj(daft2)).values / N ** 4
```
roxyboy, line 544: The errors made it past the `AttributeError`. I think we're almost at the finish line to merge this. The current error is stemming from test losing its metadata. Can you remove `.values`? I think the test should work then :)

roxyboy (changes requested):

avik2007:
Thanks - I was trying to figure out what the problem was here.

Also, I haven't been able to push a few  recent change attempts due to the
formatting with "black" failing. I also haven't been able to reload the
testing environment. Would the testing environment affect the formatting
success?

roxyboy:
> Also, I haven't been able to push a few recent change attempts due to the formatting with "black" failing. I also haven't been able to reload the testing environment. Would the testing environment affect the formatting success?

I don't think so... What's the Black error you're getting?

avik2007:
Hi Takaya,

Funny enough, I was just able to commit and push through an update.
Apparently the error is gone now. Hopefully this passes the tests.
Best,
Avik

avik2007:
@roxyboy I'm sorry I'm not able to see what the change you're requesting is.

xrft/tests/test_xrft.py
```diff
@@ -541,10 +541,11 @@ def test_cross_spectrum(self, dask):
             detrend="constant",
             window_correction=True,
         )
-        test = (daft * np.conj(daft2)).values / N ** 4
+        test = (daft * np.conj(daft2)) / N ** 4
         window, _ = _apply_window(da, dim, window_type="hann")
         dk = np.diff(np.fft.fftfreq(N, 1.0))[0]
-        test /= dk ** 2 * (window ** 2).mean()
+        test /= dk ** 2
+        test /= (window ** 2).mean()
```
roxyboy, line 548: Instead of splitting this up in two lines, maybe you could just have `/= (dk ** 2 * (window ** 2).mean())`

roxyboy:
The tests have all passed to I'm going to merge this. Thanks for your work @avik2007 ! :)

avik2007:
Thanks for all your help!

PR #179, opened 2022-04-18
Fix padding bug on no round coordinates

santisoler:
Use `numpy.linspace` to build the padded coordinate arrays instead of
`numpy.arange`. Add a new test function against coordinate arrays with no
round floats.

Fixes #172

PR #180, opened 2022-04-18
Fix typos on test_padding.py

santisoler:
Fix a few typos on docstrings and comments of test functions in `test_padding.py`


roxyboy:
Thanks for this PR in addition to #179 @santisoler ! Could you run `black` on your code to fix the formatting?
Also, should be drop support for Python 3.6...? @rabernat

santisoler:
> Could you run black on your code to fix the formatting?

Sure! Black is complaining because the stable version has changed some styles, like simple operands of power that now should be hugged (e.g. `e**a` instead of `e ** a`). I'll open a new PR for making the changes on the entire `xrft`. If we merge it and then update this one, Black shouldn't complain here.

santisoler:
@roxyboy I think this is ready to be merged now that we solved the issue with Black. Do you agree?

roxyboy:
> @roxyboy I think this is ready to be merged now that we solved the issue with Black. Do you agree?

Yes, we can merge this :)

PR #165, opened 2021-08-31
Switching from versioneer to setuptools-scm

roxyboy:
A PR to switch from `versioneer` to `setuptools_scm` for managing versions.

fixes #164

PR #182, opened 2022-04-26
Update black to 22.3.0 and ditch the deprecated pre-commit Action

### LINKED ISSUE #183: Pre-commit GitHub Action has been deprecated
The [`pre-commit/action`](https://github.com/pre-commit/action) has been deprecated, the GitHub Actions that use it are failing (see #180 and #182).

Since `xrft` uses only black on `pre-commit`, we could easily replace the `pre-commit` action by installing Black through pip and running it as a checker.

I'll be glad to tackle this issue.

santisoler:
Make `pre-commit` to use the latest stable version of Black: 22.3.0. Delete the
`pre-commit.yaml` Action workflow. Replace the deprecated `pre-commit/action`
for installing and running Black in the `linting.yaml` Action workflow.
Reformat `xrft` code with the latest Black stable release.

Fix #183


rabernat:
Thanks for this contribution @santisoler!

In order to get our linting checks to pass, you also have to update

https://github.com/xgcm/xrft/blob/42abb0fd0474f891cab84e0ede86d7f18c16b8b4/.pre-commit-config.yaml#L1-L6

@jbusecke and @roxyboy, can you explain why we have _two_ liniting github workflows (code-style and pre-commit)?

roxyboy:
> @jbusecke and @roxyboy, can you explain why we have _two_ liniting github workflows (code-style and pre-commit)?

This probably has to do with legacy configuration. I think the two linting workflows are coming from `.github/workfows` and `pre-commit-config.yml`. I don't know which one is the more up-to-date best practice... @jbusecke Do you have any take on this?

jbusecke:
As far as i can tell one was introduced by #120 and then the second one was added in #165. I think this was just an oversight? I dont see a reason to have both. They are functionally the same as far as I can see with one exception: `linting.yaml` does [run on PRs](https://github.com/xgcm/xrft/blob/42abb0fd0474f891cab84e0ede86d7f18c16b8b4/.github/workflows/linting.yaml#L1-L6), which is helpful IMO. 

I also just checked the branch [protection rules](https://github.com/xgcm/xrft/settings/branches) and saw that there are no required checks! Once these two here are consolidated I recommend setting the linting + core CI actions as required tests to be merged.

roxyboy:
> yip that would work. Not sure if you want to handle that in this PR or in a separate one though. Happy to set the branch protections if you want.

This PR is about making Black formatting work smoothly so I think it would be ok to delete it in this PR. Not sure what branch protection is but that sounds like a good idea. Please do @jbusecke  :)

roxyboy:
> After I created this PR I found there are two issues with black.
> 
> 1. The first one is the one mentioned here: Black stable version has changed some style rules, therefore the old code will fail against it.
> 2. The second one is described in [Pre-commit GitHub Action has been deprecated #183](https://github.com/xgcm/xrft/issues/183) (the `pre-commit` action has been deprecated, therefore it fails to run in Actions).
> 
> If you agree I can solve both issues in this PR.

Yes please to both :) Thanks @santisoler

jbusecke:
Actually I think we should remove black from those files. Users should always go through `pre-commit run --all-files` which will ensure the black versions between local/ci environments are the same.

jbusecke:
I have just refactored the CI environments in #184, and also removed black from there. If those tests pass, you could rebase this on on top of #184

jbusecke:
> The second one is described in https://github.com/xgcm/xrft/issues/183 (the pre-commit action has been deprecated, therefore it fails to run in Actions).

Oh wow @santisoler thanks for making me aware of this! Ill have to put in some work to update all my repos haha.

jbusecke:
> Happy to set the branch protections if you want.

I just set those, and require all python versions to pass (and also disallow pushes to main). These can be overwritten by admins (but you will get a nice red warning before).

jbusecke:
> Actually I think we should remove black from those files. Users should always go through pre-commit run --all-files which will ensure the black versions between local/ci environments are the same.

It might be best to mention this in the docs, similarly to what we have in [xgcm](https://github.com/xgcm/xgcm/blob/master/doc/contributor_guide.rst#code-formatting)

roxyboy:
> > Actually I think we should remove black from those files. Users should always go through pre-commit run --all-files which will ensure the black versions between local/ci environments are the same.
> 
> It might be best to mention this in the docs, similarly to what we have in [xgcm](https://github.com/xgcm/xgcm/blob/master/doc/contributor_guide.rst#code-formatting)

It seems that it's already mentioned [here](https://github.com/xgcm/xrft/blob/master/doc/contributor_guide.rst)...? In any case, I feel like this PR is ready to merge?

jbusecke:
Ah excellent, thanks for pointing that out @roxyboy. Let me rebase this and then when the tests pass this is good to go from my end.

roxyboy:
> Ah excellent, thanks for pointing that out @roxyboy. Let me rebase this and then when the tests pass this is good to go from my end.

Cool, thanks for your edits @jbusecke  :) Do you agree @santisoler that this PR is ready to merge?

.github/workflows/linting.yaml
```diff
@@ -1,19 +1,26 @@
...
+      - name: checkout
+        uses: actions/checkout@v2
+
+      - name: set up python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Install black
+        run: pip install black
```
jbusecke, line 23: Should we pin an explicit version of black here until the pre-commit.ci is implemented?

santisoler: I agree with @roxyboy that both the `linting.yaml` Action and `pre-commit` should be running the same version of Black. Either by pinning Black on the Action or by unpinning Black in `pre-commit` configurations. I would leave that decision to the maintainers.

santisoler:
> Cool, thanks for your edits @jbusecke :) Do you agree @santisoler that this PR is ready to merge?

Sure! Feel free to take any decision you feel the best regarding pinning or unpinning Black and merge it right away.

.github/workflows/linting.yaml
jbusecke, line 23:
```suggestion
        run: pip install black==21.6b0
```

jbusecke: I pinned the idential version as in the pre-commit config file for now. @roxyboy, if you decide to merge that suggestion, feel free to merge the PR.

santisoler: Shouldn't it be `black==22.3.0` so it installs the stable version?

jbusecke: sorry, allergies are ruining my brain these days haha.

roxyboy: Ok, Black is pinned now so I'll merge this. Thanks @santisoler and @jbusecke :)

Issue #177, opened 2022-03-28
New release


lanougue:
Hi guys,

What about a new release with all the new functions and removing all the warnings added in the last release ? ;)

rabernat:
:+1: from me!

roxyboy:
Fine with me too.

roxyboy:
I'll tidy this up after PR #182 gets merged.

roxyboy:
I'm going to close this with #186 .

PR #186, opened 2022-05-10
Linting for a new version release

roxyboy:
This PR removes the deprecated `False` flags from the functions and sets them to `True` as default. This addresses issue #177 .


roxyboy:
@lanougue Would you be ok with the linting I've done?

lanougue:
Hi @roxyboy ,
In cross_phase calculation, the default phase seems to be still "False". To be coherent, it should be changed to True, right ?

PR #190, opened 2023-01-17
Explicit error when coordinates are not numercial

lanougue:

PR #191, opened 2023-01-23
Removing deprecated call to xr.ufuncs.angle()

lanougue:
In xrft.cross_phase(), xr.ufuncs.angle() is now deprecated.
Using directly np.angle(), as suggested for replacement, do not return xarray.DataArray.
Solution is to use xr.apply_ufunc() for now.


lanougue:
Adding these modifications in #190

PR #195, opened 2023-02-10
Raise error after bad coords while padding

santisoler:
Raise a ValueError if any bad coordinate is found in the DataArray while padding. A bad coordinate is defined as any additional coordinate present in the DataArray that shares at least one of the dimensions along which the grid will be padded. The error message suggests the user to drop the offending coordinate before trying to pad the grid.


roxyboy:
This went under my radar but I think I'll merge this.

santisoler:
Thanks @roxyboy!

PR #196, opened 2023-02-22
Refactor duplicated code in xrft.py into separate functions; add one test; format with black

simonkeys:
I'm thinking of adding some features to xrft, so I started out by doing some refactoring. I added one test to check the case of ifft with unsorted coordinates.

If this is too many changes at once, I can break it up into more PRs with fewer commits.

### DIFF
--- a/xrft/tests/test_xrft.py
+++ b/xrft/tests/test_xrft.py
@@ -432,7 +432,7 @@ class TestSpectrum(object):
             ).mean("t_segment")
             # Check the amplitude correction
             # The factor of 0.5 is there because we're checking the two-sided spectrum
-            npt.assert_allclose(ps.sel(freq_t=fsig), 0.5 * A**2 / 2.0)
+            npt.assert_allclose(ps.sel(freq_t=fsig), 0.5 * A ** 2 / 2.0)
 
         da = xr.DataArray(
             np.random.rand(2, N, N),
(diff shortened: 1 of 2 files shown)


roxyboy:
Thanks a lot for this refactoring @simonkeys ! Just out of curiosity, what features are you planning to add?

simonkeys:
Basically whatever comes up in the project I'm using it for (image processing for interferometric microscopy). So far I'm thinking
- interoperability with pint-xarray
- function returning the coordinates that would appear in a fft
- possible option to back with scipy fft library instead of numpy (maybe a speed-up? I'm not sure about that)
- option to do a fft with a real dimension but then return negative and positive frequencies instead of just positive

roxyboy:
> * interoperability with pint-xarray
> * function returning the coordinates that would appear in a fft
> * possible option to back with scipy fft library instead of numpy (maybe a speed-up? I'm not sure about that)
> * option to do a fft with a real dimension but then return negative and positive frequencies instead of just positive

All features you suggest would be a great addition to `xrft`. Regarding scipy over numpy, this is something I've thought about during the early stages of developing `xrft` but do you see any clear advantages of one over the other?

roxyboy:
All the checks passed so I'm going to merge this PR. I'd be happy to continue the discussion if you could raise an issue on Github :)

PR #208, opened 2024-07-02
fix #200: raise TypeError on unused kwargs

### LINKED ISSUE #200: `xrft.fft` swallows unused kwargs and does not raise a TypeError
If I run
```python
xrft.fft(data, doesnt_exist=False)
```
I expect an TypeError to be raised since I used an `kwarg` that does not exist.

The bug is caused by lines 377ff:
```python
    if "real" in kwargs:
        real_dim = kwargs.get("real")
        warnings.warn(_real_flag_warning, FutureWarning)
```
There, you don't check whether `kwargs` is empty or not.

This bug is especially frustrating if you have a typo in a kwarg that DOES exist, like `true_amplitudes` instead of `true_amplitude`.

As far as I can tell, this only happens for `xrft.fft` and `xrft.ifft`.

TL, DR:
raise a TypeError a la `TypeError: fft() got an unexpected keyword argument 'doesnt_exist'`

cgahr:


roxyboy:
I'll merge this too :)

Issue #202, opened 2023-07-05
`xrft.pad` raises a ValueError if called with `pad_width=0`


cgahr:
If I call
```python
da.sizes  # {'x': 32}
xrft.pad(da, {'x': 0})
```
the following value error is called:
```
ValueError: could not broadcast input array from shape (0,) into shape (32,)
```

Instead, I expect the original array to be returned, as it is for `xr.pad`:
```python
da.pad({'x': 0})  # works
```

PR #209, opened 2024-07-02
fix #201 and #202

### LINKED ISSUE #201: `xrft.pad` does not work with dask array
Calling `xrft.pad(da, ...)` on a dask array result in a DataArray with slices in the frequency domain. As a result, it is not possible to perform an ifft on said array.

In my opinion, this is unwanted behavior since otherwise, `xrft` natively support parallel computing using dask.

cgahr:


roxyboy:
Thanks for the PR! Can you run black on your code for formatting purposes? I can merge this then :)

cgahr:
Could've sworn I ran black already..

roxyboy:
Merged :)

Issue #219, opened 2025-08-12
Parseval's thereom using isotropic spectrum


miniufo:
The docs have a nice section verifying the Parseval's theoreom.  But I am curious if this can hold when using isotropic spectrum.  Below is a quick example, which fails to verify the theoreom:
```python
import numpy as np
import xarray as xr
import xrft

# 2D grid size
nx, ny = 128, 128

# coordinates unit: m
x = np.linspace(0, 2*np.pi, nx) - np.pi
y = np.linspace(0, 2*np.pi, ny) - np.pi

dx = 2*np.pi / nx
dy = 2*np.pi / ny

# random a 2D field
u = xr.DataArray(np.random.randn(ny, nx), coords=[("y", y), ("x", x)])

# get the isotropic spectrum
ps = xrft.isotropic_power_spectrum(u, dim=["y", "x"], nfactor=6)

# get total energy at both spaces
E_phys = (u**2).sum(["y","x"]) * dx * dy             # \iint u^2 dx dy
E_spec = ps.sum() * (ps.freq_r[1] - ps.freq_r[0]) # \int ps(rho) d\rho

# output
print(f"Total energy in physical space: {E_phys.values}")
print(f"Total energy in spectral space: {E_spec.values}")
```

which gives (depending on the random data that may vary):
```python
Total energy in physical space: 40.02084675234615
Total energy in spectral space: 0.22568741619659388
```

So how to make the correct scaling here?

miniufo:
Hi @roxyboy, I've looked into the issue and changed `isotropize`.  Based on my understanding, current `isotropize` function takes an azimuthal average, and return radial spectrum.  If we want the total energy or variance, we could take an azimuthal `sum` (rather than `mean`), and then the returned spectrum could be summed radially to meet the Parseval's thereom.  In this way, there is no need to take care of `nbins`.

The modified `isotropize` contains two changes:
- change agg-method from `mean` to `sum`;
- when returning spectrum, we don't need to multiply `iso_ps.freq_r`

Testing codes are added to [Parseval_example.ipynb](https://github.com/xgcm/xrft/blob/master/doc/Parseval_example.ipynb).

PR is also sent, and please check on that.

This is a core function, so changing it may affect other results depending on it.

I am also curious why there is no `freq_r.spacing` in `isotropize`.  The spacing seems to be uneven.  We may need this to get a unified behavior verifying Parseval's thereom as in standard 2D x-y case.

roxyboy:
[...]

PR #222, opened 2026-02-13
change isotropize to satisfy Parseval's thereom

miniufo:
This PR tries to solve #219.

Also, we add 3D detrending codes, which may be useful when calculating 3D (2D space and 1D time) frequency-wavenumber spectrum.


roxyboy:
@miniufo I agree that this is kind of unexpected behavior but the GitHub log is saying that `test_xrft-checkpoint.py` is missing the formatting...

```
> Run black --check xrft
  black --check xrft
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.8.18/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.8.18/x64/lib
would reformat xrft/tests/.ipynb_checkpoints/test_xrft-checkpoint.py
```

miniufo:
That's great~

miniufo:
Hi @roxyboy, I've recently found that in a 2D homogeneous turbulence, the zonal/meridional spectra are not the same as isotropic spectra.  The following codes passed the test of Parseval's thereom:
```python
import xrft
import numpy.testing as npt
import xarray as xr
import numpy as np

ds = xr.open_dataset('/mnt/Data3/2DTurbulence/8192/non/sim_bih_ray_flow_part09.nc', chunks={'time':1})
psi = ds.psi[0].load()  # 8192 * 8192 streamfunction

psihx = xrft.power_spectrum(psi, dim=['x'], scaling='density').mean('y')  # x-component
psihy = xrft.power_spectrum(psi, dim=['y'], scaling='density').mean('x')  # y-component
psihi = xrft.isotropic_power_spectrum(psi, dim=['x', 'y'], scaling='density',
                                      truncate=True)                      # r-component

npt.assert_almost_equal((psi**2 * dx).sum('x').mean('y'),   # variance in grid space
                         psihx.sum('freq_x'), decimal=5)    # variance in spectral space
npt.assert_almost_equal((psi**2 * dy).sum('y').mean('x'),   # variance in grid space
                         psihy.sum('freq_y'), decimal=5)    # variance in spectral space
npt.assert_almost_equal((psi**2 * dx * dy).sum(['x', 'y']), # total variance in grid space
                         psihi.sum('freq_r'), decimal=5)    # total variance in spectral space
```

but give the plot like:
download

which is a bit surprising to me that the radial spectra differs from x-/y-components by a constant.  The constant could be `dkr / (Lx * Ly)`.  Here the important thing is the unit of the result (also determines the magnitude).  I am expecting the radial spectra is of the same magnitude as the x-/y-components.

Also, I find the [docstring](https://github.com/xgcm/xrft/blob/efc1c3006e3194acc9aed0a002415a2a14fc8353/xrft/xrft.py#L705) for `scaling='density'` or `scaling='spectrum'` is a bit unclear.  I feel the difference can be made more clearly by the units of the outputs.  I guess one is `[A]^2 m^3` while the other is `[A]^2 m^2`.  Do you have any comments on this?

roxyboy:
> Hi @roxyboy, I've recently found that in a 2D homogeneous turbulence, the zonal/meridional spectra are not the same as isotropic spectra. 

Yes, they are not the same due to the Jacobian that is needed for coordinate transformation from Cartesian to polar coordinates.

 
> Also, I find the [docstring](https://github.com/xgcm/xrft/blob/efc1c3006e3194acc9aed0a002415a2a14fc8353/xrft/xrft.py#L705) for `scaling='density'` or `scaling='spectrum'` is a bit unclear. I feel the difference can be made more clearly by the units of the outputs. I guess one is `[A]^2 m^3` while the other is `[A]^2 m^2`. Do you have any comments on this?

Regarding the `density` and `spectrum` docstrings, they are somewhat of a [(confusing but commonly accepted) jargon used in the field of FFT](https://scipy.github.io/devdocs/tutorial/signal.html#spectra-with-averaging) and we decided to adopt them.

miniufo:
Thank you for sharing these. @roxyboy.  I was trying to evaluate the spectra from a model simulation agaist SWOT observation.  The calculation based on the model outputs uses 2D isotropic spectrum while the one based on SWOT uses along-track 1D power spectrum.  Then how to reconcile these if they are not at the same energy level?

roxyboy:
I'm going to add back the option of taking the azimuthal mean in isotropizing the 2D spectra #225 . 
The summation approach is going to introduce strong isotropizing artifacts when the aspect ratio of the Fourier transformed domain is highly uneven.

PR #227, opened 2026-06-30
Extend fft and ifft to allow cupy-xarray use

### LINKED ISSUE #77: Feature Request: GPU implementation option.
The FFT implemented by numpy is pretty fast already, but for some applications it is not preferable to use a CPU (even vectorized and parallel) implementation. Ex: when most GPU cores are taken up with other concurrent dask processes.

There exist a set of GPU python-CUDA bindings provided by several packages out there, so adding the option to do rfft, irfft, fft, and ifft operations may be standardized already.

pcarlip:
Modifies the fft and ifft functions for compatibility with cupy-xarray, to allow for cuda-accelerated ffts. Resolves #77. I included at least some test coverage, but I don't know the test suite very well and I don't think I extended all tests. I also haven't tested cupy-dask combinations, which have caused problems for me elsewhere. Please let me know if this is a feature you're interested in, and if so whether it makes more sense to discuss it here or on the issue.

### DIFF
--- a/xrft/tests/test_padding.py
+++ b/xrft/tests/test_padding.py
@@ -59,7 +59,7 @@ def test_pad_coordinates_invalid(sample_da_2d):
     """
     Test if pad_coordinates raises error after unevenly spaced coords
     """
-    x = sample_da_2d.coords["x"].values
+    x = sample_da_2d.coords["x"].values.copy()
     x[3] += 0.1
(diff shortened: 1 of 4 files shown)


pcarlip:
(The first commit is unrelated, it just fixes some of the tests since directly modifying coords is not allowed with recent xarray versions)

